PR #3, opened 2022-04-20
DPE-213 add build step to CI

WRFitch:
## Issue
Currently the CI fails to run integration tests, since deploying the charm requires the pgbouncer container. This PR adds a stage to the CI that downloads and builds the necessary docker container for use in the microk8s 

## Release Notes
- Adds a stage to the CI that builds the docker container necessary for the PgBouncer k8s charm. 
- Removed leftover actions from actions.yaml


.github/workflows/ci.yaml
```diff
@@ -36,6 +65,15 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          name: image
+          path: /tmp
+      - name: Load Image into microk8s
+        run: |
+          /usr/bin/sg microk8s -c "microk8s ctr image import /tmp/image.tar"
       - name: Run integration tests
         # set a predictable model name so it can be consumed by charm-logdump-action
         run: tox -e integration -- --model itest
```
marceloneppel, line 79: I guess the problem is here. The model that is created by the actions operator is named `testing`.

marceloneppel, line 79: I guess the problem is here. The model that is created by the actions operator is named `testing`.

WRFitch: This was one of the problems; I'd also messed up the docker container build.

metadata.yaml
```diff
@@ -20,5 +26,5 @@ resources:
     type: oci-image
     description: OCI image for pgbouncer
     # currently this uses a local build of https://github.com/canonical/pgbouncer-container - see CONTRIBUTING.md for instructions.
-    upstream-source: pgbouncer:0.8
-    local-source: pgbouncer:0.8
+    upstream-source: pgbouncer:latest
```
zmraul, line 29: Is `latest` preferred over a specific version? Meaning, is upstream pgbouncer solid when it comes to new releases?

CONTRIBUTING.md
```diff
@@ -27,8 +27,8 @@ This setup is required for testing and deploying this charm. These instructions
 # Import container
 git clone https://github.com/canonical/pgbouncer-container.git
 # Build container locally, since it's not been exported anywhere yet.
-docker build . -t pgbouncer:0.8
-docker save pgbouncer:0.8 -o pgb.tar
+docker build . -t pgbouncer:latest
```
MiaAltieri, line 30: Quick question, why latest instead of a specific version.

I can imagine a scenario where a new version is realised and maybe the charm breaks because of the new version. (But also tbh I don't know much about docker and this stuff so this might be an empty concern.)

.github/workflows/ci.yaml
```diff
@@ -23,25 +23,64 @@ jobs:
...
+      - name: Build Image and Export
+        uses: docker/build-push-action@v2
+        with:
+          context: ./
+          file: ./Dockerfile
+          builder: ${{ steps.buildx.outputs.name }}
+          # Do not publish the image.
+          push: false
+          # Set the tag to retrieve the image in Deploy tests.
+          tags: pgbouncer:latest
```
MiaAltieri, line 50: Question why latest instead of a specific version, (see followup below)

README.md
```diff
@@ -13,22 +13,6 @@ As this charm is not yet published, you need to follow the build and deploy inst
...
-  - Removes an existing pgbouncer user.
-  - Params:
-    - username
-- `change-password`
-  - Updates the password for a given pgbouncer user.
-  - Params:
-    - username
-    - password
-- `get-users`
-  - Lists the existing pgbouncer users.
```
MiaAltieri, line 31: You kind of elluded to this in your description, but why are leftover actions being removed?

metadata.yaml
```diff
@@ -20,5 +26,5 @@ resources:
...
+    local-source: pgbouncer:latest
```
MiaAltieri, line 30: why latest

MiaAltieri (commented):
Looks good to me, I just had some questions.

src/charm.py
```diff
@@ -225,7 +225,7 @@ def _push_userlist(self, users: Dict = None, reload_pgbouncer: bool = False) ->
             if pgb_container.pull(USERLIST_PATH).read() == userlist:
                 logger.info("updated userlist does not modify existing pgbouncer userlist")
                 return
-        except FileNotFoundError:
+        except (FileNotFoundError, ConnectionError):
```
zmraul, line 228: Same as L166

zmraul (approved):
LGTM! Some minor things that I want to understand better :)

CONTRIBUTING.md
WRFitch: This was just a name picked at random (docker sets its most recent build to `latest` by default), but the tag should ideally the same across local and CI copies since that's what charmcraft expects. I can change this to `pgbouncer:local` or something similar to better communicate this, if that helps?

README.md
WRFitch: The handlers for these actions were removed in the last PR because this charm should be accessing postgres users from its relation to the postgres charm.

shayancanonical (approved):

paulomach (approved):
LGTM Will, previous reviews addressed my doubts

CONTRIBUTING.md
MiaAltieri: Yes!

PR #6, opened 2022-06-30
Add backend-db-admin legacy relation

WRFitch:
## Proposal
The k8s charm needs to implement the backend-db-admin legacy relation so it can connect to the backend postgres charm (or anything that provides the db-admin relation). This has been implemented as a separate module. 

## Context
- This code is copy/pasted from [the corresponding PR in the machine charm](https://github.com/canonical/pgbouncer-operator/pull/9/), with some modifications to get it to work on k8s. 
- Charm currently doesn't block without this relation, because testing this manually is time-consuming. When I implement integration tests for this relation, I'll implement this expected behaviour as well. 

## Release Notes
- Add relations/backend_db_admin.py, and corresponding unit tests. 
- Charm library updated
- Some updates to docs

## Testing
- Unit tests
- Manual testing
  - Created an offer on machine postgres charm
  - related to machine postgres charm from k8s charm
  - scaled postgres up and down, including deleting master node, monitoring the according pgbouncer config changes 
  - removed relation, monitoring that pgbouncer effectively 
- No integration tests yet, since they either require a k8s postgres charm or an automated cross-model relation. Once the postgres k8s charm is uploaded to charmhub, this will be added.


lib/charms/pgbouncer_operator/v0/pgb.py
```diff
@@ -56,13 +63,26 @@ class PgbConfig(MutableMapping):
...
+                pgb.DEFAULT_CONFIG can be used here as a default dict.
+        """
         self.__dict__.update(*args, **kwargs)
 
         if isinstance(config, str):
             self.read_string(config)
         elif isinstance(config, dict):
             self.read_dict(config)
+        elif isinstance(config, PgbConfig):
+            self.read_dict(config)
```
marceloneppel, line 85: This can be simplified to `isinstance(config, (dict, PgConfig))`.

lib/charms/pgbouncer_operator/v0/pgb.py
```diff
@@ -56,13 +63,26 @@ class PgbConfig(MutableMapping):
...
+            config: an existing config. Can be passed in as a string, dict, or PgbConfig object.
+                pgb.DEFAULT_CONFIG can be used here as a default dict.
+        """
         self.__dict__.update(*args, **kwargs)
 
         if isinstance(config, str):
             self.read_string(config)
         elif isinstance(config, dict):
             self.read_dict(config)
+        elif isinstance(config, PgbConfig):
```
shayancanonical, line 84: can probably be combined with the condition on line 82 (like it was in the other PR)

metadata.yaml
```diff
@@ -28,3 +28,22 @@ resources:
...
+    optional: true
+  db-admin:
+    interface: pgsql
+    optional: true
+
+requires:
+  # Legacy relations - these will be deprecated in a future release
+  backend-db-admin:
+    interface: pgsql
+    optional: true
```
shayancanonical, line 49: missing new line

src/relations/backend_db_admin.py
```diff
@@ -0,0 +1,198 @@
...
+
+        self._update_standbys(cfg, standbys)
+
+        self.charm._render_pgb_config(cfg, reload_pgbouncer=True)
+
+    def _on_relation_departed(self, departed_event: RelationDepartedEvent):
+        """Handle backend-db-admin-relation-departed event.
+
+        Removes unit information from pgbouncer config when a unit is removed.
+        """
```
shayancanonical, line 138: is it possible that the departing postgres unit was the master? if so, would we need to update `pg_master` key before reloading the config?

also, would we need to screen the departed_event to make sure we're noop-ing if the departing unit is a pgbouncer unit?

lib/charms/pgbouncer_operator/v0/pgb.py
```diff
@@ -27,19 +27,26 @@
 from collections.abc import MutableMapping
 from configparser import ConfigParser, ParsingError
 from copy import deepcopy
-from typing import Dict
+from typing import Dict, Union
 
 logger = logging.getLogger(__name__)
 
 PGB_DIR = "/var/lib/postgresql/pgbouncer"
+PORT_MAX = 49151  # Maximum valid port number before we get into ephemeral ports
```
shayancanonical, line 35: is this variable being used?

marceloneppel (approved):
LGTM! I'm also curious about the questions that Shayan wrote.

src/relations/backend_db_admin.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+
+        cfg = self.charm._read_pgb_config()
+        dbs = cfg["databases"]
+
+        dbs.pop("pg_master", None)
+
+        for db in list(dbs.keys()):
+            # Remove all standbys
+            if db[:21] == STANDBY_PREFIX:
```
paulomach, line 195: I didn't get why until the `21`

lib/charms/pgbouncer_operator/v0/pgb.py
WRFitch: Whoops, missed that. Thanks!

lib/charms/pgbouncer_operator/v0/pgb.py
WRFitch: This was going to be a part of how we set `listen_port` in the config, but it's no longer relevant.

paulomach (commented):
Observing @shayancanonical comments plus one of my own before approving it.

src/relations/backend_db_admin.py
WRFitch: When a postgres master is removed, the "master" field in the db-admin relation data is updated, causing a relation-changed event. The relation-changed event updates master accordingly, and should actually remove the relevant standbys, so this hook might be irrelevant - I'll look into that. 

Currently there's no pgbouncer scaling implemented (the peer relation is just a stub), so I haven't done your suggestion, but I'll add it in once I add scaling.

lib/charms/pgbouncer_operator/v0/pgb.py
WRFitch: Missed that, thanks!

src/relations/backend_db_admin.py
WRFitch: `STANDBY_PREFIX` is 21 characters long, so we're checking if `db` is a backend standby. I've replaced 21 with `len(STANDBY_PREFIX`.

paulomach (approved):
lgtm

shayancanonical (approved):
lgtm!

PR #7, opened 2022-07-11
Update backend db admin relation

WRFitch:
## Proposal
The legacy backend relation is no longer necessary, and caused bugs with user generation. Therefore, implement a new one using the data_platform_libs.database_requires lib. 

## Context
- Everything added under the `lib/` folder is a new lib being imported to this charm, and doesn't need review. 
- Integration testing could be implemented now, but I've chosen to wait until the new postgres charm is pushed to charmhub. 

## Release Notes
- Added postgresql lib
- Added data_platform_libs.database_requires lib
- Removed old backend relation, and replaced it with a newer, more lightweight relation. 
- Stored backend relation data as part of the `charm` object as `postgresql_backend`, which will be used to implement the db relations. 

## Testing
- Added unit tests
- Tested manually with new postgres charm. 
- Integration tests will be added once the postgres charm is released.


src/charm.py
```diff
@@ -110,6 +125,19 @@ def _pgbouncer_layer(self) -> Layer:
             },
         }
 
+    def _on_update_status(self, _) -> None:
+        """Update Status hook.
+
+        Uses systemd status to verify pgbouncer is running.
```
marceloneppel, line 131: It is using the availability of the backend relation, right?

src/charm.py
```diff
@@ -204,6 +236,92 @@ def _reload_pgbouncer(self) -> None:
...
+        Args:
+            user: the username for the intended user
+            cfg: A PgbConfig object. Modified during this method.
+            password: intended password for the user
+            admin: whether or not the user has admin permissions
+            stats: whether or not the user has stats permissions
+            reload_pgbouncer: whether or not to restart pgbouncer after changing config. Must be
+                restarted for changes to take effect.
+            render_cfg: whether or not to render config
+        Returns False if user cannot be created.
```
marceloneppel, line 262: Maybe it's better to return an exception to handle in the method that calls this method? I'm thinking about that because there is no return value for success and also there is no value being returned in some cases (like when the user is already created in the files).

src/charm.py
```diff
@@ -204,6 +236,92 @@ def _reload_pgbouncer(self) -> None:
...
+
+        Args:
+            user: the username for the intended user.
+            cfg: A PgbConfig object. Modified during this method.
+            reload_pgbouncer: whether or not to restart pgbouncer after changing config. Must be
+                restarted for changes to take effect.
+            render_cfg: whether or not to render config
+
+        Returns:
+            False if user creation fails.
```
marceloneppel, line 306: Something to think about like the above comment.

src/relations/backend_database.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Postgres backend-db-admin relation hooks & helpers.
+
+This relation uses the pgsql interface.
+
+Some example relation data is below. The only parts of this we actually need are the "master" and
+"standbys" fields. All values are examples taken from a test deployment, and are not definite.
```
marceloneppel, line 9:
```suggestion
Some example relation data is below. The only parts of this we actually need are the "endpoints" and
"read-only-endpoints" fields. All values are examples taken from a test deployment, and are not definite.
```

marceloneppel (approved):
LGTM!

paulomach (approved):
lgtm

src/charm.py
WRFitch: Whoops, copy-paste error. You're right, pebble should be handling whether pgbouncer is running, so it's more about checking we're connected.

src/charm.py
WRFitch: You're right, I'll add a new exception in. The charm.py files are maybe getting a little big, but I'm planning to do some refactors once these relations are done.

WRFitch: I've actually updated this to use a FileNotFoundError based on our mattermost chat.  There's not a great place to put a dedicated exception with the charm as it exists now, but there's going to be some refactors in the future once the relations are sorted and the charm is released, so I'll add your suggestion then.

marceloneppel (approved):

PR #10, opened 2022-07-20
Add db relations

WRFitch:
## Proposal
Adds legacy `db` and `db-admin` relations to pgbouncer. 

## Context
- listen_port needs to be added to config because some charms expect the default postgres port of 5432, without any option to change it. 
- I apologise for the extreme hugeness of this PR

## Release Notes
- Adds db and db-admin relations, and associated integration tests. 
- Added listen_port to config

## Testing
- Unit tests
- Integration tests are adapted from postgres charm to ensure we present the same interface. 
  - Finos-Waltz test tests `db` relation, as well as verifying config changes update relations. 
  - Discourse test tests `db-admin` relation
- manually tested pgbouncer with discourse charm


tests/unit/relations/test_db.py
```diff
@@ -0,0 +1,390 @@
...
+        standbys = self.db_relation._get_standbys(cfg, app, cfg_entry, db_name, user, pw)
+        standby_list = standbys.split(", ")
+
+        assert len(standby_list) == 2
+
+        assert cfg["databases"][f"{cfg_entry}_standby"] == {
+            "host": "host2",
+            "dbname": db_name,
+            "port": "port2",
+        }
```
marceloneppel, line 330: Why just checking the second endpoint? Shouldn't it be the `{cfg_entry}_standby_1` key?

src/relations/db.py
```diff
@@ -0,0 +1,325 @@
...
+        try:
+            external_unit = self.get_external_units(change_event.relation)[0]
+        except IndexError:
+            # In cases where pgbouncer changes the relation, we have no new information to add to
+            # the config. Scaling is not yet implemented, and calling this hook from the
+            # backend-db-admin relation occurs after the config updates are added.
+            logger.info(
+                f"no external unit found in {self.relation_name} relation - nothing to change in config, exiting relation hook"
+            )
+            return
```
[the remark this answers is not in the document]
WRFitch: PGB has only triggered its own events when I've told it to; when the backend-database relation updates, the new connection information has to be written to the config, so it calls the config-changed hook for both `db` relations if they exist. Outside of this use case, I don't think this is a problem.

tests/integration/relations/test_db_admin.py
```diff
@@ -0,0 +1,164 @@
...
+APPLICATION_UNITS = 1
+DATABASE_UNITS = 3
+
+logger = logging.getLogger(__name__)
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+PGB = METADATA["name"]
+PG = "postgresql-k8s"
+PSQL = "psql"
+APPS = [PG, PGB, PSQL]
```
marceloneppel, line 34: Are PSQL or APPS supposed to be used in some part of the test?

src/relations/db.py
```diff
@@ -0,0 +1,414 @@
...
+        )
+        self.framework.observe(
+            charm.on[self.relation_name].relation_broken, self._on_relation_broken
+        )
+
+        self.charm = charm
+        self.admin = admin
+
+    def _on_relation_joined(self, join_event: RelationJoinedEvent):
+        """Handle db-relation-joined event."""
```
marceloneppel, line 124: You could add a summary to the method description explaining what is done in this handler like you did in the relation changed handler.

src/relations/db.py
```diff
@@ -0,0 +1,414 @@
...
+            self.charm.backend_postgres.create_database(database, user)
+
+            created_msg = f"database and user for {self.relation_name} relation created"
+            self.charm.unit.status = ActiveStatus(created_msg)
+            logger.info(created_msg)
+        except (PostgreSQLCreateDatabaseError, PostgreSQLCreateUserError):
+            err_msg = f"failed to create database or user for {self.relation_name}"
+            logger.error(err_msg)
+            self.charm.unit.status = BlockedStatus(err_msg)
+            join_event.defer()
```
marceloneppel, line 192: Maybe it is better to not defer the event? I am thinking about some error that is not transient.

src/relations/db.py
```diff
@@ -0,0 +1,414 @@
...
+
+        cfg = self.charm.read_pgb_config()
+        user = app_databag["user"]
+        database = app_databag["database"]
+
+        # delete user
+        self.charm.remove_user(user, cfg=cfg, render_cfg=True, reload_pgbouncer=True)
+        try:
+            logging.error(self.charm.backend_postgres.delete_user)
+            self.charm.backend_postgres.delete_user(user, if_exists=True)
```
marceloneppel, line 357: I believe we need to add a check to ensure the backend PostgreSQL relation is still active (otherwise PostgreSQL would have removed the pgbouncer user and this can fail; it will be handled also in some way in the PostgreSQL side later, to delete users that PgBouncer created if the relation is broken between them).

marceloneppel (commented):
Looks great, Will! I left some comments to discuss some details.

tests/integration/relations/helpers/helpers.py
```diff
@@ -0,0 +1,79 @@
...
+        "--container",
+        "pgbouncer",
+        "pgbouncer-k8s-operator/0",
+        "cat",
+        f"{pgb.PGB_DIR}/pgbouncer.ini",
+    )
+    return pgb.PgbConfig(cat_cfg[1])
+
+
+def wait_for_relation_joined_between(ops_test: OpsTest, endpoint_one: str, endpoint_two: str):
```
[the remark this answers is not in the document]
WRFitch: Thanks! :D

tests/unit/relations/test_db.py
WRFitch: the db relations only expect (and use) one relation, named `f"{database}_standby" - this has been updated to match the new version.

README.md
```diff
@@ -10,15 +10,21 @@ As this charm is not yet published, you need to follow the build and deploy inst
 
 ## Relations
 
+Please only create relations on charms that are in active-idle status. Removing relations does not delete data.
```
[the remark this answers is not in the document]
WRFitch: I think the backend_postgres check & deferral in `_on_relation_changed` should handle this situation. I'll remove this line.

README.md
```diff
@@ -27,7 +33,7 @@ The following relations provide support for the [LMA charm bundle](https://juju.
 
 ## OCI Images
 
-This charm uses the canonical pgbouncer-container docker image, available [here](https://github.com/canonical/pgbouncer-container). As this container has not been uploaded anywhere, **you will need to build this image locally and import it into your container registry before use**, following the instructions in [CONTRIBUTING.md]([CONTRIBUTING.md](https://github.com/canonical/pgbouncer-k8s-operator/CONTRIBUTING.md)).
+This charm uses the canonical pgbouncer-container docker image, available [here](https://github.com/canonical/pgbouncer-container), stored in dockerhub.
```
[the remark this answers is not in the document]
WRFitch: So that's already been updated in metadata.yaml, but I've updated the docs to have a link to dockerhub.

src/charm.py
```diff
@@ -82,14 +87,20 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         """Handle changes in configuration."""
         container = self.unit.get_container(PGB)
         if not container.can_connect():
-            self.unit.status = WaitingStatus("waiting for pgbouncer workload container.")
+            container_err_msg = "waiting for pgbouncer workload container."
+            logger.error(container_err_msg)
```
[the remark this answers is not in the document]
WRFitch: I've updated this one to be logger.info, and the others are warnings. Is that ok?

tests/integration/relations/test_db.py
```diff
@@ -0,0 +1,166 @@
...
+            apps=[ANOTHER_FINOS_WALTZ],
+            status="blocked",
+            raise_on_blocked=False,
+            timeout=1000,
+        )
+
+        cfg = await get_cfg(ops_test)
+        logger.info(cfg)
+        logger.info(await pgbouncer_app.get_config())
+        assert cfg["pgbouncer"]["listen_port"] == port
```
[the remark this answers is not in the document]
WRFitch: Yeah it's not really necessary anymore, I'll get rid of it

tests/integration/relations/test_db_admin.py
WRFitch: nope, removed

src/relations/db.py
WRFitch: This is a good idea, I'll implement this.

src/relations/db.py
WRFitch: good idea!

src/relations/db.py
WRFitch: You're right - this was originally written as a catch if backend_postgres wasn't finished implementing but we don't start the hook unless it's ready to go, so I'll remove the deferral.

src/relations/db.py
WRFitch: So this was a little more complicated than I thought, but I've added two checks: one checks at the beginning of the hook that the backend and db relations haven't been initialised, so if this hook somehow gets called before relation_joined (due to the deferral queue getting a bit weird), we should wait for the relation to initialise, then immediately remove. I'd love to just remove the relation entirely, but since we can't control the hook queue I think this is the best I can do. 

The second check verifies backend_postgres when it's used, when we're deleting user. If it's inaccessible at this point, we assume the backend relation has been removed, and postgres will remove the users pgbouncer has created, so we skip deleting the user. Every method that deletes users shouldn't raise errors if the user in question isn't deleted, so operation should continue even if the users are being deleted in some other way. 

Does this work for you?

marceloneppel: Yeah It works! Thanks Will!

marceloneppel (approved):

README.md
```diff
@@ -27,7 +31,7 @@ The following relations provide support for the [LMA charm bundle](https://juju.
 
 ## OCI Images
 
-This charm uses the canonical pgbouncer-container docker image, available [here](https://github.com/canonical/pgbouncer-container). As this container has not been uploaded anywhere, **you will need to build this image locally and import it into your container registry before use**, following the instructions in [CONTRIBUTING.md]([CONTRIBUTING.md](https://github.com/canonical/pgbouncer-k8s-operator/CONTRIBUTING.md)).
+This charm uses the canonical pgbouncer-container docker image, available [here](https://github.com/canonical/pgbouncer-container), stored in [dockerhub](https://hub.docker.com/r/dataplatformoci/pgbouncer).
```
paulomach, line 34: The repo should be `https://code.launchpad.net/~data-platform/+git/pgbouncer`

requirements.txt
```diff
@@ -1,2 +1,2 @@
 ops >= 1.5.0
-tenacity==8.0.1
\ No newline at end of file
+tenacity==8.0.1
```
paulomach, line 2: nit: new line

src/relations/db.py
```diff
@@ -0,0 +1,440 @@
...
+        # postgres application because we don't want to delete all user data with one command.
+        delete_db = False
+        for relname in ["db", "db-admin"]:
+            for relation in self.charm.model.relations.get(relname, []):
+                if relation.id == broken_event.relation.id:
+                    continue
+                if relation.data.get(self.charm.app, {}).get("database") == database:
+                    # There's multiple applications using this database, so don't remove it until
+                    # we can guarantee this is the last one.
+                    delete_db = True
```
marceloneppel, line 385: Should it really be deleted if there are more than one relation using it? The above comment says that we shouldn't delete the database.

WRFitch: Good catch, this has been updated

src/relations/backend_database.py
```diff
@@ -100,6 +105,9 @@ def _on_relation_broken(self, event: RelationBrokenEvent):
             cfg = self.charm.read_pgb_config()
             user = self.charm.backend_postgres.user
             self.charm.remove_user(user, cfg=cfg, reload_pgbouncer=True, render_cfg=True)
+
+            self.charm.update_postgres_endpoints()
```
marceloneppel, line 109: This is not updating (removing) the PostgreSQL endpoints in the pgbouncer config file. As the endpoints are still present in the databag during this event, they are kept in the pgbouncer config file.

marceloneppel (approved):
LGTM! Great job on simplifying the events, Will! You can consider my comments for a new PR. This one is fine.

paulomach (approved):
lgtm - I second Marcelo's comments though

src/charm.py
```diff
@@ -447,19 +449,29 @@ def backend_postgres(self) -> PostgreSQL:
 
         return PostgreSQL(host=host, user=user, password=password, database=database)
 
-    def trigger_db_relations(self):
-        """Triggers every instance of client relations db and db-admin, if they exist."""
-        # Skip triggering relations if backend_postgres doesn't exist yet.
+    def update_backend_relation_port(self, port):
+        """Update ports in backend relations to match updated pgbouncer port."""
```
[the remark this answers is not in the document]
WRFitch: yep, totally right. I'll update that in the next PR

PR #11, opened 2022-08-05
transferring lib ownership to k8s charm

WRFitch:
## Proposal
The pgbouncer charm lib previously lived on the VM charm. Since I'm planning to use the k8s charm for development moving forward, porting implementation over to the VM charm once it's fully implemented, tested, and reviewed, it makes sense to keep the charm library on the k8s charm as well. 

## Context
- I'm planning to use the k8s charm for development because it builds and deploys faster, making testing and development faster. 
- All this code has already been reviewed on the VM charm, I'm just changing the names of things and copying unit tests over. 

## Release Notes
- Rename pgb lib and copy unit tests over from VM charm.

## Testing
- Unit and integration tests have run


MiaAltieri (approved):
LGTM

zmraul (changes requested):
Hm, shouldn't the change be `/pgbouncer_k8s/` instead of `/pgbouncer_k8s_operator/`? I think that's the way most charms are setting up libs.

WRFitch:
> Hm, shouldn't the change be `/pgbouncer_k8s/` instead of `/pgbouncer_k8s_operator/`? I think that's the way most charms are setting up libs.

I believe that's because the charm is named `pgbouncer_k8s` or similar, and I haven't removed the redundant `_operator` part. I'll do that now, though.

paulomach (approved):
lgtm - loved the rational behind moving the lib to k8s and will consider it for my stuff

marceloneppel (approved):
LGTM!

zmraul (approved):

PR #12, opened 2022-08-10
Remove userlist

WRFitch:
## Proposal
The existing implementation authenticates users based on a list of usernames and plaintext passwords in a text file in the default user's root directory. This PR changes authentication to query the postgres database for user passwords. 

## Context
-  The SQL scripts are copied from [percona's docker postgres implementation](https://github.com/percona/percona-docker/tree/main/postgresql-containers/bin/postgres-ha/sql/pgbouncer), with some small modifications. 
  - This uses the Apache2 license, so as long as we still give credit, legally we're all good. 
  - The function has to be generated in its own schema in each database so there's no single function that can access everything. 
- Unfortunately we have to retain the userlist file to authenticate the auth_user, but the password is hashed using md5
- pgbouncer auth docs: 
  - https://www.pgbouncer.org/config.html#authentication-file-format
  - https://www.pgbouncer.org/config.html#authentication-settings
- Changes to database_requires.py are imported from the lib

## Release Notes
- Updated user authentication to query postgres, rather than checking a plaintext file 
  - Created dedicated auth_user and auth_function, which runs in an automatically generated schema for each database requested by client relations. 
- Moved user management to pgb library, with some minor qol updates 
- Tightened up file permissions
- moved install hook to a start hook. 

## Testing
- Integration and unit tests have been updated


tests/unit/relations/test_db.py
```diff
@@ -26,40 +30,59 @@ def setUp(self):
...
+        # Define a db relation
+        self.db_rel_id = self.harness.add_relation(DB_RELATION_NAME, "postgres")
+        self.harness.add_relation_unit(self.db_rel_id, "client/0")
+        self.harness.add_relation_unit(self.db_rel_id, self.unit)
+
+        # Define a db-admin relation
+        self.db_admin_rel_id = self.harness.add_relation(DB_ADMIN_RELATION_NAME, "postgres")
+        self.harness.add_relation_unit(self.db_admin_rel_id, "admin_client/0")
+        self.harness.add_relation_unit(self.db_admin_rel_id, self.unit)
+
```
marceloneppel, line 54: You can change `postgres`to another application name here in both cases. I think it makes more sense.

src/charm.py
```diff
@@ -144,7 +127,8 @@ def _pgbouncer_layer(self) -> Layer:
                 PGB: {
                     "summary": "pgbouncer service",
                     "user": PG_USER,
-                    "command": f"pgbouncer {INI_PATH}",
+                    # -R flag reuses sockets on restart
+                    "command": f"pgbouncer -R -v {INI_PATH}",
```
marceloneppel, line 131: Great addition!

src/relations/backend_database.py
```diff
@@ -69,45 +81,166 @@ def __init__(self, charm: CharmBase):
...
+        databag = self.app_databag
+        endpoint = databag.get("endpoints")
+        user = databag.get("username")
+        password = databag.get("password")
+        database = self.relation.data[self.charm.app].get("database")
+
+        if None in [endpoint, user, password, database]:
+            return None
+
+        if endpoint is not None:
```
marceloneppel, line 219: You can remove this check as it it also being verified in the above check.

src/relations/backend_database.py
```diff
@@ -69,45 +81,166 @@ def __init__(self, charm: CharmBase):
...
+            postgres = self.postgres
+
+        install_script = open("src/relations/pgbouncer-install.sql", "r").read()
+
+        with postgres.connect_to_database(dbname) as conn, conn.cursor() as cursor:
+            cursor.execute(install_script.replace("auth_user", self.auth_user))
+        conn.close()
+        logger.info("auth function initialised")
+
+    def get_postgres(self, host, user, password, database) -> PostgreSQL:
```
marceloneppel, line 185: I guess you can move the code from this method to the `postgres` property below and use that in the places this one is used, as the event data that is being passed to this method is from the event that comes from PostgreSQL, which is also in the app databag that is used in the `postgres` property. This suggestion is optional, as I can be missing something in the details of this method use.

marceloneppel (approved):
Great job, Will! I left some comments about suggestions. Feel free to consider or not each of them.

tests/integration/relations/test_backend_database.py
```diff
@@ -57,34 +59,39 @@ async def test_create_backend_db_admin_legacy_relation(ops_test: OpsTest):
...
+                        and "auth_query" not in cfg["pgbouncer"].keys()
                     ):
                         break
         except RetryError:
-            assert False, "pgbouncer config files failed to update in 3 minutes "
+            assert False, "pgbouncer config files failed to update in 3 minutes"
+
+        cfg = await get_cfg(ops_test)
+        logging.info(cfg.render())
+        logger.info(await get_pgb_log(ops_test))
```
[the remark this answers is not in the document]
WRFitch: This just prints the pgbouncer log to the test output - it's not always obvious what's going on from pgb's perspective, so this logging makes it a bit easier to debug without having to catch flaws as they happen.

src/relations/backend_database.py
WRFitch: Agreed, this has been

src/charm.py
WRFitch: Thanks!

marceloneppel (approved):

src/relations/backend_database.py
```diff
@@ -69,45 +81,152 @@ def __init__(self, charm: CharmBase):
...
+            postgres: a PostgreSQL application instance. If none is provided, the default pgbouncer
+                instance will be used.
+            dbname: the name of the database to connect to.
+        """
+        logger.info("initialising auth function")
+
+        install_script = open("src/relations/pgbouncer-install.sql", "r").read()
+
+        with self.postgres.connect_to_database(dbname) as conn, conn.cursor() as cursor:
+            cursor.execute(install_script.replace("auth_user", self.auth_user))
```
MiaAltieri, line 175: is there any error checking that should be done?

src/relations/backend_database.py
```diff
@@ -69,45 +81,152 @@ def __init__(self, charm: CharmBase):
...
+        conn.close()
+        logger.info("auth function initialised")
+
+    @property
+    def relation(self) -> Relation:
+        """Relation object for postgres backend-database relation."""
+        backend_relation = self.model.get_relation(RELATION_NAME)
```
MiaAltieri, line 182: I believe you may need to specify a relation id (if there are multiple relations of the same name you'll need an id to guarantee you get the right one)

(or is it the case that PGB only relates to one PG?)

src/relations/db.py
```diff
@@ -135,7 +134,7 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
         If the backend relation is fully initialised and available, we generate the proposed
         database and create a user on the postgres charm, and add preliminary data to the databag.
         """
-        if not self.charm.backend_postgres:
+        if not self.charm.backend.postgres:
```
MiaAltieri, line 137: might you need to check if PBG is also ready and running?

src/relations/db.py
```diff
@@ -414,23 +427,13 @@ def _on_relation_broken(self, broken_event: RelationBrokenEvent):
...
-                # removed and will handle user deletion in its own relation-broken hook.
-                self.charm.backend_postgres.delete_user(user, if_exists=True)
-        except PostgreSQLDeleteUserError:
-            blockedmsg = f"failed to delete user for {self.relation_name}"
-            logger.error(blockedmsg)
-            self.charm.unit.status = BlockedStatus(blockedmsg)
-            return
+        cfg.remove_user(user)
+        self.charm.render_pgb_config(cfg, reload_pgbouncer=True)
+        self.charm.backend.postgres.delete_user(user)
```
MiaAltieri, line 436: should there be some error handling here?

src/relations/pgbouncer-install.sql
```diff
@@ -0,0 +1,77 @@
...
+    pg_authid.rolname = $1;
+$$
+LANGUAGE SQL STABLE SECURITY DEFINER;
+
+/**
+ * As mentioned, the pgbouncer user will only be able to access its one function
+ * and all it can do is execute. Here is where it does exactly that
+ */
+REVOKE ALL ON FUNCTION auth_user.get_auth(username TEXT) FROM PUBLIC, auth_user;
+GRANT EXECUTE ON FUNCTION auth_user.get_auth(username TEXT) TO auth_user;
```
MiaAltieri, line 77: this is something I can't review so might be good to ask someone who is more familiar with SQL

src/relations/pgbouncer-uninstall.sql
```diff
@@ -0,0 +1,47 @@
...
+ * Drop anything owned by the pgbouncer user. It should be nothing at this
+ * point, but better safe than sorry...
+ */
+DROP OWNED BY auth_user CASCADE;
+
+/**
+ * So, we can't drop the pgbouncer role as this file runs on an individual
+ * database. We need **all** objects associated "pgbouncer" dropped before we
+ * can execute `DROP ROLE pgbouncer;`
+ */
```
MiaAltieri, line 47: this is something I can't review so might be good to ask someone who is more familiar with SQL

tests/integration/relations/test_db_admin.py
```diff
@@ -49,6 +50,7 @@ async def test_create_db_admin_legacy_relation(ops_test: OpsTest):
                 resources=resources,
                 application_name=PGB,
             ),
+            # TODO test only deploying 2 postgres instances
```
MiaAltieri, line 53: ?

tests/integration/relations/test_db_admin.py
```diff
@@ -163,3 +167,5 @@ async def test_create_db_admin_legacy_relation(ops_test: OpsTest):
             pg_user=pgb_user,
             pg_user_password=pgb_password,
         )
+
+        logger.info(await get_pgb_log(ops_test))
```
MiaAltieri, line 171:
```suggestion
```

tests/integration/relations/test_db.py
```diff
@@ -163,15 +160,12 @@ async def test_create_db_legacy_relation(ops_test: OpsTest):
...
-
         cfg = await get_cfg(ops_test)
         logger.info(cfg)
         assert finos_user not in cfg["pgbouncer"]["admin_users"]
         assert another_finos_user not in cfg["pgbouncer"]["admin_users"]
 
         assert "waltz" not in cfg["databases"].keys()
         assert "waltz_standby" not in cfg["databases"].keys()
+
+        logger.info(await get_pgb_log(ops_test))
```
MiaAltieri, line 171:
```suggestion
```

tests/integration/relations/test_backend_database.py
```diff
@@ -57,34 +59,39 @@ async def test_create_backend_db_admin_legacy_relation(ops_test: OpsTest):
         wait_for_relation_joined_between(ops_test, PG, PGB)
         await ops_test.model.wait_for_idle(apps=[PGB, PG], status="active", timeout=1000),
 
-        userlist = await get_userlist(ops_test)
         cfg = await get_cfg(ops_test)
-        pgb_user = f"relation_id_{relation.id}"
-        pgb_password = userlist[pgb_user]
+        logging.info(cfg.render())
```
MiaAltieri, line 63:
```suggestion
```

tests/integration/relations/test_backend_database.py
```diff
@@ -57,34 +59,39 @@ async def test_create_backend_db_admin_legacy_relation(ops_test: OpsTest):
...
+        assert cfg["pgbouncer"]["auth_query"]
 
         await check_database_users_existence(ops_test, [pgb_user], [], pgb_user, pgb_password)
 
         # Remove relation but keep pg application because we're going to need it for future tests.
         await ops_test.model.applications[PG].remove_relation(
             f"{PGB}:{RELATION}", f"{PG}:database"
         )
+        pgb_unit = ops_test.model.applications[PGB].units[0]
+        logging.info(await get_app_relation_databag(ops_test, pgb_unit.name, relation.id))
```
MiaAltieri, line 75: is this meant to test something? If not testing maybe remove it

tests/integration/relations/helpers/helpers.py
```diff
@@ -69,30 +72,36 @@ async def get_app_relation_databag(ops_test: OpsTest, unit_name: str, relation_i
...
         "ssh",
         "--container",
         "pgbouncer",
         f"{PGB}/0",
         "cat",
-        f"{pgb.PGB_DIR}/pgbouncer.ini",
+        filepath,
     )
-    return pgb.PgbConfig(cat_cfg[1])
+    return cat[1]
```
MiaAltieri, line 104:
```    
    cat_cmd = f"ssh -- container pgbouncer {PGB}/0 cat {pgb.PGB_DIR}/pgbouncer.ini} "
    return_code, output, _ = await ops_test.juju(*cat_cmd.split())

    if return_code != 0:
        raise ProcessError(
            "Expected kill command %s to succeed instead it failed: %s", kill_cmd, return_code
        )

   return output
```

tests/unit/relations/test_db.py
```diff
@@ -26,40 +30,61 @@ def setUp(self):
         self.harness.begin()
 
         self.charm = self.harness.charm
+        self.app = self.charm.app.name
+        self.unit = self.charm.unit.name
+        self.backend = self.charm.backend
         self.db_relation = self.charm.legacy_db_relation
         self.db_admin_relation = self.charm.legacy_db_admin_relation
 
+        # TODO update mocks now we're using the mock harness
```
MiaAltieri, line 39: remove? or which mocks are you referring to

src/relations/backend_database.py
```diff
@@ -69,45 +81,152 @@ def __init__(self, charm: CharmBase):
...
+
+        plaintext_password = pgb.generate_password()
+        # create authentication user on postgres database, so we can authenticate other users
+        # later on
+        self.postgres.create_user(self.auth_user, plaintext_password, admin=True)
+        self.initialise_auth_function(dbname=self.database.database)
+
+        hashed_password = pgb.get_hashed_password(self.auth_user, plaintext_password)
+        self.charm.push_file(
+            f"{PGB_DIR}/userlist.txt", f'"{self.auth_user}" "{hashed_password}"', perms=0o777
```
WRFitch, line 114: TODO update these permissions

lib/charms/pgbouncer_k8s/v0/pgb.py
```diff
@@ -170,19 +173,14 @@ def _parse_complex_variables(self) -> None:
...
             raise PgbConfig.ConfigParsingError(source=str(err))
 
-        try:
-            for name, cfg_string in self[users].items():
-                self[users][name] = parse_kv_string_to_dict(cfg_string)
-        except KeyError:
-            # [users] section is not compulsory, so continue.
-            pass
+        for name, cfg_string in self.get(users, dict()).items():
+            self[users][name] = parse_kv_string_to_dict(cfg_string)
```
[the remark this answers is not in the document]
WRFitch: we're using self.get() instead with a default value, so there's no chance of a KeyError.

tests/integration/relations/test_db_admin.py
WRFitch: These logging lines are actually there intentionally, so we can see what pgbouncer is doing when something goes wrong - it makes the test logs a bit longer, but I think it makes them more useful.

tests/integration/relations/test_db_admin.py
WRFitch: whoops, remove

lib/charms/pgbouncer_k8s/v0/pgb.py
```diff
@@ -170,19 +173,14 @@ def _parse_complex_variables(self) -> None:
...
 
         for user_type in PgbConfig.user_types:
-            try:
-                self[pgb][user_type] = self[pgb][user_type].split(",")
-            except KeyError:
-                # user type fields are not compulsory, so continue
-                pass
+            users = self[pgb].get(user_type, "").split(",")
+            if "" in users:
+                users.remove("")
```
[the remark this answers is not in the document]
WRFitch: good point, I'll update this to cast it to a set before removing empty users.

src/charm.py
```diff
@@ -77,32 +68,24 @@ def _on_install(self, event: InstallEvent) -> None:
...
-        self._render_userlist(pgb.initialise_userlist_from_ini(config))
+        self.render_pgb_config(config)
 
     def _on_config_changed(self, event: ConfigChangedEvent) -> None:
         """Handle changes in configuration."""
         container = self.unit.get_container(PGB)
         if not container.can_connect():
             container_err_msg = "waiting for pgbouncer workload container."
-            logger.info(container_err_msg)
+            logger.debug(container_err_msg)
```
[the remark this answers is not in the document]
WRFitch: It's not really an error, so I've updated the variable name to `wait_msg

src/relations/backend_database.py
```diff
@@ -69,45 +81,152 @@ def __init__(self, charm: CharmBase):
...
+        # adds user to pgb config
+        cfg.add_user(user=event.username, admin=True)
+        cfg["pgbouncer"][
+            "auth_query"
+        ] = f"SELECT username, password FROM {self.auth_user}.get_auth($1)"
+        cfg["pgbouncer"]["auth_file"] = f"{PGB_DIR}/userlist.txt"
+        self.charm.render_pgb_config(cfg)
+
+        self.charm.update_postgres_endpoints(reload_pgbouncer=True)
+
```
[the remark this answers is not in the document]
WRFitch: Nope, they should only fire if there's relation objects to run them and all they're doing is updating pgb config and databags.

src/relations/backend_database.py
```diff
@@ -69,45 +81,152 @@ def __init__(self, charm: CharmBase):
...
+    ):
+        self.charm.update_postgres_endpoints(reload_pgbouncer=True)
+
+    def _on_relation_departed(self, event: RelationDepartedEvent):
+        if event.departing_unit != self.charm.unit:
+            return
+
+        logger.info("removing auth user")
+
+        uninstall_script = open("src/relations/pgbouncer-uninstall.sql", "r").read()
```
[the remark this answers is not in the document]
WRFitch: The uninstall script doesn't remove the auth_user, I've just added a line in to do that and am testing it now. The uninstall script basically just removes all of the permissions from the auth_user before it's deleted - I'll think of a better way to document what it's doing

src/relations/backend_database.py
```diff
@@ -69,45 +81,152 @@ def __init__(self, charm: CharmBase):
...
-        try:
-            cfg = self.charm.read_pgb_config()
-            user = self.charm.backend_postgres.user
-            self.charm.remove_user(user, cfg=cfg, reload_pgbouncer=True, render_cfg=True)
+        cfg = self.charm.read_pgb_config()
+        cfg.remove_user(self.postgres.user)
+        cfg["pgbouncer"].pop("auth_user", None)
+        cfg["pgbouncer"].pop("auth_query", None)
+        self.charm.render_pgb_config(cfg)
+        self.charm.delete_file(f"{PGB_DIR}/userlist.txt")
```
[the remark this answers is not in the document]
WRFitch: Pgbouncer is only tested with one postgres relation, and the intention is that it's only ever deployed that way, so managing multiple backend-database relations isn't necessary.

src/relations/backend_database.py
```diff
@@ -69,45 +81,152 @@ def __init__(self, charm: CharmBase):
...
+
+    def _on_relation_departed(self, event: RelationDepartedEvent):
+        if event.departing_unit != self.charm.unit:
+            return
+
+        logger.info("removing auth user")
+
+        uninstall_script = open("src/relations/pgbouncer-uninstall.sql", "r").read()
+
+        with self.postgres.connect_to_database(PGB_DB) as conn, conn.cursor() as cursor:
```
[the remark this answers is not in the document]
WRFitch: If we can't connect to database here, then we can't really fix anything, so it's hard to figure out what _to_ do. I've set a BlockedStatus and an early return, but I'm open to ideas.

src/relations/backend_database.py
WRFitch: PGB should only ever relate to one postgres instance

src/relations/backend_database.py
WRFitch: This function is only being used in relation_joined hooks, where we're deferring if self.postgres isn't accessible. Running this basically anywhere else is a bad idea, so I'm happy to let this error out, but I'll document how and why it works.

tests/unit/relations/test_db.py
WRFitch: removed :)

src/relations/db.py
WRFitch: PGB doesn't actually have to be running for this relation to come online, but it does need to have a valid config file, so I've added a check for that.

src/relations/db.py
WRFitch: We're already checking backend.postgres exists in line 416, so I'm not sure this is necessary. If we can't access it despite having the credentials, I'd argue that's an issue in postgres.

tests/integration/relations/test_backend_database.py
WRFitch: Nope, more logging so the state of the application at this point is easily observable.

tests/integration/relations/helpers/helpers.py
WRFitch: nice!

src/relations/backend_database.py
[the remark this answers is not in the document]
WRFitch: I've added this into the docstring for the function  :)

src/relations/backend_database.py
```diff
@@ -69,45 +82,166 @@ def __init__(self, charm: CharmBase):
...
 
-            self.charm.update_postgres_endpoints()
+        self.postgres.delete_user(self.auth_user)
 
-        except FileNotFoundError:
-            event.defer()
+        logger.info("auth user removed")
 
-    def _on_relation_broken(self, event: RelationBrokenEvent):
+    def _on_relation_broken(self, _: RelationBrokenEvent):
```
paulomach, line 160: If you are not caring about the event, you should not care about it's type either, i.e.:
```suggestion
    def _on_relation_broken(self, _):
```

lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         self.password = password
         self.database = database
 
-    def _connect_to_database(self, database: str = None) -> psycopg2.extensions.connection:
+    def connect_to_database(self, database: str = None) -> psycopg2.extensions.connection:
```
paulomach, line 69: I guess you need to bump this lib version, right?

paulomach (approved):
lgtm

lib/charms/postgresql_k8s/v0/postgresql.py
marceloneppel: We are going to bump that lib version in the PostgreSQL k8s charm. There is also some changes that were made in the lib through the PostgreSQL VM charm that we will need to consolidate.

marceloneppel (approved):
LGTM!

MiaAltieri (approved):

PR #14, opened 2022-08-12
Add lib to charmhub

WRFitch:
## Proposal
Adds charmhub variables to pgb lib so it can be published, thus allowing the CI to run properly.


paulomach (approved):
exquisite

marceloneppel (approved):
Good idea!

PR #16, opened 2022-08-24
Scaling

WRFitch:
## Proposal
This PR allows PgBouncer to scale to multiple units. 

## Context
- We're storing everything in the peer databag for now, including usernames and passwords for the db relations. These will be removed to Juju secrets once they exist. 
- The config is stored as an ini file in the databag, rather than a json file, because it's an existing string representation of the data and it's more readable (although there's a bug in jhack that doesn't display the headers)

## Release Notes
- added pgb_peers relation and corresponding integration test run
- updated integration test runners to use juju version 2.9.29 to work around scale-down bug. 
- added juju get_secret interface

## Testing
- updated unit tests
- updated the existing db integration test to work with scaled pgbouncer
  - db-admin test has not been updated because discourse behaves weirdly


tests/integration/relations/test_backend_database.py
```diff
@@ -105,6 +107,41 @@ async def test_pgbouncer_stable_when_deleting_postgres(ops_test: OpsTest):
...
-        await ops_test.model.applications[PG].remove()
-        wait_for_relation_removed_between(ops_test, PG, PGB)
-        await ops_test.model.wait_for_idle(apps=[PGB], status="active", timeout=1000)
+        await scale_application(ops_test, PGB, 3)
+        await asyncio.gather(
+            ops_test.model.wait_for_idle(
+                apps=[PGB], status="active", timeout=1000, wait_for_exact_units=3
+            ),
+            ops_test.model.wait_for_idle(
+                apps=[PG], status="active", timeout=1000, wait_for_exact_units=3
```
marceloneppel, line 116: Just out of curiosity, do you know if `wait_for_exact_units` works if we provide multiple apps in the same `wait_for_idle` call?

tests/integration/relations/test_backend_database.py
```diff
@@ -105,6 +107,41 @@ async def test_pgbouncer_stable_when_deleting_postgres(ops_test: OpsTest):
...
+            ),
+        )
+
+        username = f"relation_id_{relation.id}"
```
marceloneppel, line 120: A reminder for me and you: I need to fix the k8s charm to create the username following [another pattern](https://github.com/canonical/postgresql-operator/pull/15#discussion_r940469355) and it should be changed later here too.

src/relations/peers.py
```diff
@@ -0,0 +1,255 @@
...
+    def app_databag(self):
+        """Returns the app databag for the Peer relation."""
+        peer_relation = self.model.get_relation(PEER_RELATION_NAME)
+        if peer_relation is None:
+            return {}
+        return peer_relation.data[self.charm.app]
+
+    @property
+    def unit_databag(self):
+        """Returns the app databag for the Peer relation."""
```
marceloneppel, line 100:
```suggestion
        """Returns the unit databag for the Peer relation."""
```

src/relations/peers.py
```diff
@@ -0,0 +1,255 @@
...
+
+        if cfg is not None or auth_file is not None:
+            try:
+                # returns an error if this is fired before on_pebble_ready.
+                self.charm.reload_pgbouncer()
+            except ConnectionError:
+                # TODO find a better error to use
+                event.defer()
+
+    def set_secret(self, scope:str, key:str, value:str):
```
marceloneppel, line 156: Wouldn't be better to move this to `charm.py` or another place, as it should be changed to use Juju secrets instead of relation data in the future?

src/relations/peers.py
```diff
@@ -0,0 +1,255 @@
...
+            # peer relation not yet initialised
+            # TODO fail louder
+            return
+
+        self.app_databag[CFG_FILE_DATABAG_KEY] = cfg.render()
+        logger.debug("updated config file in peer databag")
+
+    def get_cfg(self) -> PgbConfig:
+        """Retrieves the pgbouncer config from the peer databag."""
+        if cfg := self.get_secret("app", CFG_FILE_DATABAG_KEY):
```
marceloneppel, line 230: Is this supposed to be a secret? If so, you can also use `set_secret` on L225 and L142.

src/relations/db.py
```diff
@@ -430,7 +393,61 @@ def _on_relation_broken(self, broken_event: RelationBrokenEvent):
...
+
+        for databag in self.get_databags(relation):
+            databag.update(updates)
+
+    def _generate_username(self, event):
+        """Generates a unique username for this relation."""
+        app_name = self.charm.app.name
+        relation_id = event.relation.id
+        model_name = self.model.name
+        return f"{app_name}_user_id_{relation_id}_{model_name}".replace("-", "_")
```
marceloneppel, line 414: We can try to remove some things like the `id` text from this pattern like commented [here](https://github.com/canonical/postgresql-operator/pull/15#discussion_r940469355). I am still thinking if we can improve it a little bit more (but still thinking in cross model relations).

src/relations/peers.py
```diff
@@ -0,0 +1,255 @@
...
+            return
+
+        self.update_cfg(cfg)
+
+        try:
+            if self.charm.backend:
+                # update userlist only if backend relation exists.
+                self.update_auth_file(self.charm.read_auth_file())
+        except FileNotFoundError:
+            # Auth files will be added to the databag once they're available
```
marceloneppel, line 125: Don't you need to defer or trigger the event again in some place to populate the `AUTH_FILE_DATABAG_KEY` in the databag? Sorry if I am missing it in another place (I checked some calls and couldn't find it).

marceloneppel (commented):
Awesome work, Will!

tests/unit/relations/test_db.py
```diff
@@ -36,6 +40,10 @@ def setUp(self):
         self.db_relation = self.charm.legacy_db_relation
         self.db_admin_relation = self.charm.legacy_db_admin_relation
 
+        # Define a peer relation
+        self.peers_rel_id = self.harness.add_relation(PEER_RELATION_NAME, "postgres")
```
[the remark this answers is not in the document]
WRFitch: good catch

tests/integration/relations/test_backend_database.py
WRFitch: I think it does, but I believe it waits for the _total_ units - in this case, six. I think this is cleaner, since we know we want 3 units of PGB, and 3 of PG.

tests/unit/test_charm.py
```diff
@@ -126,6 +131,12 @@ def test_read_pgb_config(self):
 
     @patch("ops.model.Container.restart")
     def test_reload_pgbouncer(self, _restart):
-        self.charm._reload_pgbouncer()
+        self.harness.set_leader(True)
+        # necessary hooks before we can check reloads
+        self.harness.charm.on.start.emit()
+        container = self.harness.model.unit.get_container(PGB)
+        self.charm.on.pgbouncer_pebble_ready.emit(container)
```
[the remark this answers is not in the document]
WRFitch: Unfortunately we need the pebble_ready event, which doesn't seem to fire during begin_with_initial_hooks()

src/relations/peers.py
WRFitch: I think in future you're right, but I want to keep updates for this databag mostly in this file. Would a wrapper method in charm.py work?

src/relations/peers.py
WRFitch: when the auth_file is added it should create a relation_changed event. This is called in render_auth_file, in charm.py line 290. If anything, the config deferral should be unnecessary - I'm testing that now.

paulomach (approved):
lgtm - left some comments here and there, nothing significant though

src/relations/peers.py
marceloneppel: No worries. It can be done in the future.

tests/integration/relations/test_backend_database.py
marceloneppel: Thanks for the information!

tests/unit/test_charm.py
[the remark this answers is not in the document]
marceloneppel: Hmm, good to know!

src/relations/peers.py
marceloneppel: Thanks for the clarification, Will.

marceloneppel (approved):
LGTM!

src/charm.py
```diff
@@ -156,30 +160,35 @@ def _on_pgbouncer_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
+            # pebble_ready event hasn't fired so pgbouncer has not been added to pebble config
+            # TODO update error handling across the charm so we aren't constantly swapping
+            # deferred hooks around
+            raise ConnectionError
 
-        if reload_pgbouncer:
-            self._reload_pgbouncer()
+        self.unit.status = MaintenanceStatus("Reloading Pgbouncer")
+        logger.info("reloading pgbouncer application")
+        pgb_container.restart(PGB)
```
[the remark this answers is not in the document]
WRFitch: After testing this, I've added a function `check_pgb_running` to guarantee this is happening.

src/relations/backend_database.py
```diff
@@ -89,17 +88,22 @@ def __init__(self, charm: CharmBase):
...
         """Handle backend-database-database-created event.
 
         Accesses user and password generated by the postgres charm and adds a user.
         """
         logger.info("initialising postgres and pgbouncer relations")
-        if self.postgres is None:
+        if not self.charm.unit.is_leader():
+            return
+
+        if self.postgres is None or self.relation.data[self.charm.app].get("database") is None:
```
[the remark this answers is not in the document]
WRFitch: `BACKEND_RELATION_NAME` is actually "backend-database" - it's "database" from the postgres side, so the intention is for that to be something like our `CLIENT_RELATION_NAME` in future for our new relation

src/charm.py
```diff
@@ -15,18 +21,15 @@
 from ops.framework import StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
-from ops.pebble import Layer, PathError
+from ops.pebble import ConnectionError, Layer, PathError
 
+from constants import AUTH_FILE_PATH, INI_PATH, PG_USER, PGB
```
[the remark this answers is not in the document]
WRFitch: it's the pgbouncer service name, so it's `"pgbouncer"`, whereas the app name is `"pgbouncer-k8s"`.

marceloneppel (approved):

zmraul (approved):
LGTM! A couple comments

src/relations/peers.py
```diff
@@ -0,0 +1,248 @@
...
+        self.framework.observe(charm.on[PEER_RELATION_NAME].relation_created, self._on_created)
+        self.framework.observe(charm.on[PEER_RELATION_NAME].relation_joined, self._on_changed)
+        self.framework.observe(charm.on[PEER_RELATION_NAME].relation_changed, self._on_changed)
+
+    @property
+    def app_databag(self):
+        """Returns the app databag for the Peer relation."""
+        peer_relation = self.model.get_relation(PEER_RELATION_NAME)
+        if peer_relation is None:
+            return {}
```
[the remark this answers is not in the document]
WRFitch: Good catch, I thought these were necessary because of my convoluted hooks but tests seem to run fine after deleting them.

marceloneppel (approved):

PR #18, opened 2022-09-19
Fix endpoints to point to correct units when scaling

WRFitch:
## Proposal
Fixes endpoints to point to the correct units when scaling. 

## Context 
I forgot to update how master/standby endpoints are updated when adding scaling. This PR adds that in. 

## Testing
Updated unit tests, integration tests all pass fine.

### DIFF
--- a/src/relations/backend_database.py
+++ b/src/relations/backend_database.py
@@ -146,6 +146,7 @@ class BackendDatabaseRequires(Object):
             self.charm.unit.status = BlockedStatus(
                 "failed to remove auth user when disconnecting from postgres application."
             )
+            event.fail()
             return
 
         self.postgres.delete_user(self.auth_user)
--- a/src/relations/backend_database.py
+++ b/src/relations/backend_database.py
@@ -143,10 +143,9 @@ class BackendDatabaseRequires(Object):
                 cursor.execute(uninstall_script.replace("auth_user", self.auth_user))
             conn.close()
         except psycopg2.Error:
-            self.charm.unit.status = BlockedStatus(
-                "failed to remove auth user when disconnecting from postgres application."
-            )
-            event.fail()
+            auth_fail = "failed to remove auth user when disconnecting from postgres application."
+            self.charm.unit.status = BlockedStatus(auth_fail)
+            event.fail(auth_fail)
             return
 
         self.postgres.delete_user(self.auth_user)
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -353,6 +353,16 @@ class DbProvides(Object):
         Removes relevant information from pgbouncer config when db relation is removed. This
         function assumes that relation databags are destroyed when the relation itself is removed.
         """
+        # Set a flag to avoid deleting database users when this unit
+        # is removed and receives relation broken events from related applications.
+        # This is needed because of https://bugs.launchpad.net/juju/+bug/1979811.
+        # Neither peer relation data nor stored state are good solutions,
+        # just a temporary solution.
+        if departed_event.departing_unit == self.charm.unit:
+            self.charm.peers.unit_databag.update({"departing": "True"})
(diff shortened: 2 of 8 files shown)


src/relations/peers.py
```diff
@@ -88,17 +90,53 @@ def __init__(self, charm: CharmBase):
...
+    @property
+    def units_hostnames(self) -> Set[str]:
+        """Fetch current list of peers hostnames.
+
+        Returns:
+            A list of peers addresses (strings).
+        """
+        # Get all members IPs and remove the current unit IP from the list.
+        addresses = {self._get_unit_hostname(unit) for unit in self.relation.units}
+        addresses.add(self.charm.unit_pod_hostname)
```
marceloneppel, line 121: The comment says `remove` but an address is being added.

marceloneppel (dismissed):
LGTM!

src/relations/backend_database.py
```diff
@@ -146,6 +146,7 @@ def _on_relation_departed(self, event: RelationDepartedEvent):
             self.charm.unit.status = BlockedStatus(
                 "failed to remove auth user when disconnecting from postgres application."
             )
+            event.fail()
```
MiaAltieri, line 149:
```suggestion
            event.fail("failed to remove auth user when disconnecting from postgres application.")
```

src/relations/db.py
```diff
@@ -289,22 +289,27 @@ def update_port(self, relation: Relation, port: str):
...
-                "host": self.charm.unit_pod_hostname,
-                "dbname": database,
-                "port": port,
-                "user": user,
-                "password": password,
-                "fallback_application_name": self.get_external_app(relation).name,
-            }
-        )
+        master_dbconnstr = {
+            "host": self.charm.peers.leader_hostname,
```
MiaAltieri, line 293: Sorry I don't know enough about PGB & PG, is it gaurenteed that the leader is the master?

Also nit at some point you'll want to replace master with something else, see [this convo](https://chat.canonical.com/canonical/pl/kb7k7pnhbifyb8mqoubg8zb1ky), feel free to do this throughout in a separate PR.

src/relations/peers.py
```diff
@@ -121,6 +159,8 @@ def _on_created(self, event: RelationCreatedEvent):
 
     def _on_changed(self, event: RelationChangedEvent):
         """If the current unit is a follower, write updated config and auth files to filesystem."""
+        self.unit_databag[ADDRESS_KEY] = self.charm.unit_pod_hostname
```
MiaAltieri, line 162: might be good to set this in the `install` hook as well, otherwise it might only get initialised when receiving `on_changed` events.

src/relations/peers.py
```diff
@@ -141,10 +182,9 @@ def _on_changed(self, event: RelationChangedEvent):
 
         if cfg is not None or auth_file is not None:
             try:
-                # returns an error if this is fired before on_pebble_ready.
+                # raises an error if this is fired before on_pebble_ready.
                 self.charm.reload_pgbouncer()
             except ConnectionError:
-                # TODO find a better error to use
```
MiaAltieri, line 147: might be helpful to log error

MiaAltieri (approved):
LGTM some suggestions/questions/nits

src/relations/db.py
WRFitch: For pgbouncer, the leader/master distinction is irrelevant. Unfortunately the relation interface relies on this outdated terminology, so we're forced to use it

src/relations/peers.py
WRFitch: `on_changed` is pretty common, but I'll add it to the `on_joined` hook anyway :)

marceloneppel (approved):

marceloneppel (approved):

MiaAltieri (approved):

PR #19, opened 2022-09-28
Dev/add client relation

WRFitch:


tests/integration/relations/pgbouncer_provider/application-charm/src/charm.py
```diff
@@ -0,0 +1,199 @@
...
+        event.set_results({"results": json.dumps(results)})
+
+    def connect_to_database(
+        self, database: str, user: str, host: str, password: str, port: str
+    ) -> psycopg2.extensions.connection:
+        """Creates a connection to the database.
+
+        Args:
+            database: database to connect to (defaults to the database
+                provided when the object for this class was created).
```
marceloneppel, line 186: It seems that there isn't a default value for this field. It requires the one that is passed to this method.

tests/integration/relations/pgbouncer_provider/application-charm/src/charm.py
```diff
@@ -0,0 +1,199 @@
...
+        # Retrieve the credentials using the charm library.
+        logger.info(f"cluster2 credentials: {event.username} {event.password}")
+        self.unit.status = ActiveStatus("received database credentials for cluster2")
+
+    def _on_cluster2_endpoints_changed(self, event: DatabaseEndpointsChangedEvent) -> None:
+        """Event triggered when the read/write endpoints of the database change."""
+        logger.info(f"cluster2 endpoints have been changed to: {event.endpoints}")
+
+    def _on_run_sql_action(self, event: ActionEvent):
+        logger.error(event.params)
```
marceloneppel, line 145: The same for other `logger.error` calls on this file (the ones that are being used for debugging).
```suggestion
```

tests/integration/relations/pgbouncer_provider/application-charm/src/charm.py
```diff
@@ -0,0 +1,199 @@
...
+        relation_id = event.params["relation-id"]
+        databag = self.first_database.fetch_relation_data()[relation_id]
+        logger.error(databag)
+
+        dbname = event.params["dbname"]
+        query = event.params["query"]
+        user = databag.get("username")
+        password = databag.get("password")
+        if event.params["readonly"]:
+            host = databag.get("read-only-endpoints").split(",")[0]
```
marceloneppel, line 155: I think this is optional, as both endpoint are the same (what changes is just the database name, as described on README).

tests/integration/relations/pgbouncer_provider/application-charm/src/charm.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+
+    def _on_cluster_endpoints_changed(self, event: DatabaseEndpointsChangedEvent) -> None:
+        """Event triggered when the read/write endpoints of the database change."""
+        logger.info(
+            f"cluster {event.relation.app.name} endpoints have been changed to: {event.endpoints}"
+        )
+
+    # Multiple database clusters events observers (for aliased clusters/relations).
+    def _on_cluster1_database_created(self, event: DatabaseCreatedEvent) -> None:
```
marceloneppel, line 124: This and the other methods below are not needed, as they are not mentioned as callbacks for any events that are being observed.

tests/integration/relations/pgbouncer_provider/application-charm/metadata.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+containers:
+  pgbouncer:
+    resource: application-image
+
+resources:
+  application-image:
+    type: oci-image
+    description: OCI image for test charm
+    upstream-source: ubuntu:latest
```
marceloneppel, line 18: Is this needed for something? Or just to have a better structure, like a normal k8s charm?

tests/integration/relations/pgbouncer_provider/helpers.py
```diff
@@ -0,0 +1,183 @@
...
+        return False
+
+
+async def get_application_relation_data(
+    ops_test: OpsTest,
+    application_name: str,
+    relation_name: str,
+    key: str,
+    relation_id: str = None,
+    relation_alias: str = None,
```
marceloneppel, line 65: The `relation_alias` parameter can be removed as we're not testing aliases.

tests/integration/relations/pgbouncer_provider/helpers.py
```diff
@@ -0,0 +1,183 @@
...
+    read_only_endpoint: bool = False,
+) -> str:
+    """Build a PostgreSQL connection string.
+
+    Args:
+        ops_test: The ops test framework instance
+        application_name: The name of the application
+        relation_name: name of the relation to get connection data from
+        relation_id: id of the relation to get connection data from
+        relation_alias: alias of the relation (like a connection name)
```
marceloneppel, line 150: The same as above.

tests/integration/relations/pgbouncer_provider/helpers.py
```diff
@@ -0,0 +1,183 @@
...
+        ops_test,
+        application_name,
+        relation_name,
+        "read-only-endpoints" if read_only_endpoint else "endpoints",
+        relation_id,
+        relation_alias,
+    )
+    host = endpoints.split(",")[0].split(":")[0]
+
+    # Translate the service hostname to an IP address.
```
marceloneppel, line 176:
```suggestion
    # Translate the pod hostname to an IP address.
```

tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
```diff
@@ -0,0 +1,339 @@
...
+    drop_query = "DROP TABLE test;"
+    run_drop_query_readonly = await run_sql_on_application_charm(
+        ops_test,
+        unit_name=client_unit_name,
+        query=drop_query,
+        dbname=dbname,
+        relation_id=relation.id,
+        readonly=True,
+    )
+    assert "no results to fetch" in json.loads(run_drop_query_readonly["results"])
```
marceloneppel, line 144: Wouldn't be better to have an error being caught by, for example, `pytest.raises`, as the output `no results to fetch` is already returned when some queries succeed (like the ones below about creating database and user)?

tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
```diff
@@ -0,0 +1,339 @@
...
+    # and wait for them exchanging some connection data.
+    first_cluster_relation = await ops_test.model.add_relation(
+        f"{CLIENT_APP_NAME}:{MULTIPLE_DATABASE_CLUSTERS_RELATION_NAME}", PGB
+    )
+    second_cluster_relation = await ops_test.model.add_relation(
+        f"{CLIENT_APP_NAME}:{MULTIPLE_DATABASE_CLUSTERS_RELATION_NAME}", PGB_2
+    )
+    await ops_test.model.wait_for_idle(status="active")
+
+    # Retrieve the connection string to both database clusters using the relation aliases
```
marceloneppel, line 241:
```suggestion
    # Retrieve the connection string to both database clusters using the relation ids
```

src/relations/pgbouncer_provider.py
```diff
@@ -0,0 +1,265 @@
...
+
+        # Delete the user.
+        try:
+            self.charm.backend.postgres.delete_user(user)
+        except PostgreSQLDeleteUserError as e:
+            logger.exception(e)
+            self.charm.unit.status = BlockedStatus(
+                f"Failed to delete user during {self.relation_name} relation broken event"
+            )
+            raise
```
marceloneppel, line 160: Shouldn't you move the call about deleting the user from the database to before updating the config? So you can avoid raising the error again and just keep the blocked status (without changing anything in the pgbouncer config).

marceloneppel (commented):
Amazing implementation, Will!

tests/integration/helpers/helpers.py
```diff
@@ -181,3 +191,41 @@ async def scale_application(ops_test: OpsTest, application_name: str, scale: int
...
+
+
+# # TODO update this to match the old one perfectly.
+# async def build_connection_string(ops_test: OpsTest, readonly: bool = False):
+#     pgb_unit = ops_test.model.applications[PGB].units[0].name
+#     pgb_unit_address = await get_unit_address(ops_test, pgb_unit, app_name="pgbouncer-k8s")
+#     backend_relation = get_backend_relation(ops_test)
+#     pgb_app_databag = await get_app_relation_databag(
+#         ops_test, unit_name=pgb_unit, relation_id=backend_relation.id
+#     )
```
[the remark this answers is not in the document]
WRFitch: nope, removed

tests/integration/relations/pgbouncer_provider/application-charm/src/charm.py
WRFitch: You're right, but I think it's still worth verifying that the readonly endpoints are legitimate and usable.

tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
WRFitch: Because this is running an action, it's actually a bit tricky to expect specific errors with `pytest.raises`, since there's no error from the test's perspective. To get around this, I've rewritten the action to return the error in its `results` field, and it's now checking that we're throwing the correct errors for the correct reason.

tests/integration/relations/pgbouncer_provider/application-charm/src/charm.py
marceloneppel: I agree. You can keep them.

tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
```diff
@@ -315,6 +319,25 @@ async def test_read_only_endpoint_in_scaled_up_cluster(ops_test: OpsTest):
         )
 
 
+@pytest.mark.client_relation
+async def test_legacy_relation_compatibility(ops_test: OpsTest):
```
marceloneppel, line 323: Does this test exist to check that both legacy and modern relations can be established in the same pgbouncer application?

marceloneppel (dismissed):
LGTM! Thanks for the updates in the code, Will! I have one more question about one test that was added.

tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
WRFitch: Yep. Happy to delete this test if this functionality is unnecessary, but otherwise I figured it was worth testing :)

marceloneppel: I think it's good to have it. Should you update the phrase `Using legacy relations with normal relations is currently not tested.` on README as you're testing it?

taurus-forever (approved):
:-)

marceloneppel (approved):

PR #24, opened 2022-11-09
Persist userlist and PGB config on controller restart 

WRFitch:
## Proposal
When we restart the k8s controller on which a PGB operator runs, we lose our containers and therefore our config. This PR checks the peer databag for config files if they aren't available in the filesystem. 

## Context
- We can't keep logs because they aren't stored anywhere, and if they were, they should probably live in COS. 

## Release Notes
- Check peer databag for config files if they aren't available in filesystem. 

## Testing
- Added an integration test that restarts the juju controller pod.


src/charm.py
```diff
@@ -59,13 +59,23 @@ def _on_start(self, event: StartEvent) -> None:
             event.defer()
             return
 
-        if (default_config := self.peers.get_cfg()) is None:
+        try:
+            # Try and get pgb config. If it already exists in the filesystem it'll be refreshed,
+            # and if it only exists in the peer databag, pull it and write it to filesystem.
+            config = self.read_pgb_config()
```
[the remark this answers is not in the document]
WRFitch: Agreed, I've updated this comment to make more sense

src/charm.py
```diff
@@ -269,11 +279,18 @@ def read_pgb_config(self) -> PgbConfig:
             PgbConfig object containing pgbouncer config.
```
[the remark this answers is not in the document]
WRFitch: I've kept `read_pgb_config` and `peers.get_cfg` separate, so this is no longer necessary

src/relations/peers.py
```diff
@@ -289,6 +289,13 @@ def update_auth_file(self, auth_file: str) -> None:
         self.set_secret("app", AUTH_FILE_DATABAG_KEY, auth_file)
         logger.debug("updated auth file in peer databag")
 
+    def get_auth_file(self) -> str:
```
marceloneppel, line 292: Is this code being used in any part of the other files? I couldn't find its usage.

marceloneppel (commented):
Does killing the controller pod (to force it to be recreated) break the test run?

tox.ini
```diff
@@ -68,7 +68,7 @@ commands =
 description = Run all integration tests
 deps =
     pytest
-    juju
+    juju==2.9.11 # juju 3.0.0 has issues with retrieving action results
```
taurus-forever, line 71: JFMI, why do we lock such an old juju version? 2.9.37 is the latest as for today.

taurus-forever (approved):
I have single question as JFMI only. LGTM. Tnx!

tox.ini
WRFitch: This is the last 2.9 version of _libjuju_, which doesn't follow the same version as juju :)

marceloneppel (approved):

marceloneppel (dismissed):

paulomach (approved):
Looks sane and remind me that we need to take a look at cases like this on mysqland

marceloneppel (approved):

PR #25, opened 2022-11-17
Add group to pebble layer.

WRFitch:
## Proposal
adds user group when running pgbouncer through pebble

## Testing
- Updated unit test to match.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -120,6 +120,7 @@ class PgBouncerK8sCharm(CharmBase):
                 PGB: {
                     "summary": "pgbouncer service",
                     "user": PG_USER,
+                    "group": PG_USER,
                     # -R flag reuses sockets on restart
                     "command": f"pgbouncer -R {INI_PATH}",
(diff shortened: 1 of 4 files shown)


src/charm.py
```diff
@@ -120,6 +120,7 @@ def _pgbouncer_layer(self) -> Layer:
                 PGB: {
                     "summary": "pgbouncer service",
                     "user": PG_USER,
+                    "group": PG_USER,
```
taurus-forever, line 123: LGTM. Nit: I would add constant PG_GROUP to avoid questions and confusion in the future :-)

taurus-forever (dismissed):
LGTM

marceloneppel (dismissed):

src/charm.py
taurus-forever: Thank you for addressing it!

taurus-forever (approved):
LGTM. I like it! (c)

marceloneppel (dismissed):

PR #27, opened 2022-12-15
more aggressive testing on client relations

WRFitch:
## Proposal
Client relations could use a little more testing, so I've added a test to verify we can connect multiple pgbouncer deployments to a single postgres database, and added a function to quickly check the relations are functional throughout testing. 

## Release Notes
- Added a test to verify multiple pgbouncer deployments can connect to one postgres database
- added smoke check integration test function 
- Worked around a bug where pgbouncer would route connections to a nonexistent leader after it has been deleted, but before a new one has been elected
  - This cannot be fixed because of two problems in juju:
    - Scale-down bug means the leader doesn't receive its own relation-departed event, so it doesn't know when to remove itself as an endpoint. Non-leader units can't update this relation endpoint, since it's in the app databag. 
    - Leaders are not constant - despite [what the documentation says](https://juju.is/docs/sdk/leadership), electing new leaders isn't instantaneous. If it were, this period of pointing the client application to an outdated endpoint would be updated immediately, and this issue would be moot.

## Testing
The bug found in this PR was extremely transient, and could only be found when the leader unit happened to not be pgbouncer unit 0. This was tested exhaustively both locally and on CI, and after four consecutive valid runs, we believe it's solved.


tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
```diff
@@ -362,30 +330,74 @@ async def test_legacy_relation_compatibility(ops_test: OpsTest):
...
         relation_id=client_relation.id,
-        dbname=TEST_DBNAME,
+        dbname=APPLICATION_FIRST_DBNAME,
+        relation_name=FIRST_DATABASE_RELATION_NAME,
+    )
+
+
+@pytest.mark.client_relation
+async def test_multiple_pgb_can_connect_to_one_backend(ops_test: OpsTest, pgb_charm):
+    pgb_secondary = f"{PGB}-multi-backend"
```
marceloneppel, line 340: Maybe change `pgbouncer-k8s-multi-backend` to something like `pgbouncer-k8s-b` or even `secondary-pgbouncer-k8s` to avoid making people reading this code think that this PgBouncer application has multiple backends.

marceloneppel (approved):
Super! LGTM!

dragomirp (dismissed):

dragomirp (dismissed):

marceloneppel (approved):

dragomirp (approved):

PR #31, opened 2023-01-05
added jammy image

WRFitch:
## Release Notes
- Update to use new jammy container. 

## Context
- This PR isn't being merged until the tests are updated in the bundle repo, since they're currently failing and I don't want to make the fix more complicated. 

## Testing
Integration tests are currently running.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -44,7 +44,6 @@ jobs:
       - name: Upload charm to charmhub
         uses: canonical/charming-actions/upload-charm@2.1.0
         with:
-          resource-overrides: "pgbouncer-image:3"
           credentials: "${{ secrets.CHARMHUB_TOKEN }}"
           github-token: "${{ secrets.GITHUB_TOKEN }}"
           channel: "${{ steps.channel.outputs.name }}"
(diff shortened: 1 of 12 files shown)


taurus-forever (commented):
LGTM if new pgbouncer version works for us.
Up to Marcelo to decide!

.github/workflows/release.yaml
```diff
@@ -44,7 +44,6 @@ jobs:
       - name: Upload charm to charmhub
         uses: canonical/charming-actions/upload-charm@2.1.0
         with:
-          resource-overrides: "pgbouncer-image:3"
```
marceloneppel, line 47: Is [this](https://github.com/canonical/pgbouncer-k8s-operator/actions/runs/3846492286/jobs/6553180944#step:4:23) still a problem?

marceloneppel (approved):
LGTM! Just take a look on the issue with the image when publishing to Charmhub.

.github/workflows/release.yaml
WRFitch: I think overriding pgbouncer-image is unnecessary, since we can just pull it from dockerhub

taurus-forever (dismissed):
LGTM, please double check Marcelo comment about `resource-overrides: "pgbouncer-image:3"` (after merging).

marceloneppel (approved):

taurus-forever (commented):
After merging review: LGTM! Tnx!

PR #33, opened 2023-01-11
Multiple pgbouncer application instances in pebble

WRFitch:
## Proposal
This PR uses pebble to implement multiple pgbouncer instances. 

## Context
Pgbouncer is single-threaded. The accepted way to make use of this is to have multiple pgbouncer instances - one per core. On kubernetes, this could also be mitigated by only deploying containers with one CPU core.

## Release Notes
- Implements multiple pgbouncer instances, one per core on each unit.

## Testing
- Added an integrated test that verifies the instances are all running as expected.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -44,7 +44,6 @@ jobs:
       - name: Upload charm to charmhub
         uses: canonical/charming-actions/upload-charm@2.1.0
         with:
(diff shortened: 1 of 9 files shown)


marceloneppel (approved):
Awesome Will! LGTM!

marceloneppel (approved):

dragomirp (approved):

PR #50, opened 2023-03-07
All dep groups set to optional

dragomirp:
## Proposal
* Poetry groups get installed in tox environments that don't need them
* Mark all dependency groups as optional, so they get installed only when pulled explicitly

## Context
* Added linting for the charm library

## Release Notes
* Housekeeping
* Fixed typo in pgb common charm lib


taurus-forever (approved):
Tnx!

dragomirp:
Added linting for the charm lib and bumped libjuju

marceloneppel (approved):

taurus-forever (approved):
LGTM

PR #51, opened 2023-03-10
Update agent pin

dragomirp:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,8 +86,7 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: microk8s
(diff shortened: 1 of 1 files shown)


taurus-forever (approved):
Interesting idea to slowly move on the latest juju-agent (where it is possible).

marceloneppel (approved):
Great idea!

dragomirp:
> FYI it's possible to add the version to the tox-environments component of the matrix so that you don't need to duplicate the job
Reworked as you suggested

taurus-forever (approved):
Nice! Great hint Carl!

marceloneppel (approved):
Cool! Thanks for the hint Carl, and for the implementation, Dragomir!

PR #57, opened 2023-03-21
Renovate tweaks

dragomirp:
## Proposal
* Set the Renovate bot PR reviewers explicitly, so we don't have to set them manually
* Pin exact versions of packages, so that we can see the explicit current version in `pyproject.toml`
* Regex manager to autoupdate binary dependencies in `charmcraft.yaml`

## Testing
* Tested on a personal repo


charmcraft.yaml
```diff
@@ -17,4 +17,5 @@ parts:
       - libssl-dev
       - rustc
       - cargo
-    charm-binary-python-packages: [psycopg2-binary==2.9.5]
+    charm-binary-python-packages:
+      - psycopg2-binary==2.9.5  # renovate
```
dragomirp, line 21: For easier regexing

renovate.json
```diff
@@ -20,6 +21,9 @@
       "allowedVersions": "<3.0.0",
       "extractVersion": "^juju-(?<version>.*)$",
       "groupName": "Juju agent"
+    }, {
+      "matchPackageNames": ["juju"],
+      "allowedVersions": "<3.0.0"
     }
```
dragomirp, line 27: So python-libjuju isn't updated beyond 2.9 even if explicitly pinned.

renovate.json
```diff
@@ -3,14 +3,15 @@
...
+  "reviewers": ["dragomirp", "marceloneppel", "taurus-forever"],
   "enabledManagers": ["poetry", "pip_requirements", "github-actions", "regex"],
   "schedule": ["after 1am and before 2am every weekday"],
   "timezone": "Etc/UTC",
   "prHourlyLimit": 0,
   "packageRules": [
     {
-      "matchManagers": ["poetry", "pip_requirements"],
+      "matchManagers": ["poetry", "pip_requirements", "regex"],
+      "matchDatasources": ["pypi"],
```
dragomirp, line 14: So that charmcraft deps are updated together with the other python packages.

tox.ini
```diff
@@ -50,7 +50,6 @@ commands =
 description = Run unit tests
 commands =
     poetry install --with unit
-    poetry export -f requirements.txt -o requirements.txt
```
dragomirp, line 53: Renovate will update the lock and requirements, so we don't really need to export everywhere.

renovate.json
```diff
@@ -3,14 +3,15 @@
   "extends": [
     "config:base"
   ],
-  "reviewers": ["team:data-platform-postgresql"],
+  "reviewers": ["dragomirp", "marceloneppel", "taurus-forever"],
```
dragomirp, line 6: I think this is more useful than just assigning the team.

pyproject.toml
```diff
@@ -13,55 +13,55 @@ repository = "https://github.com/canonical/pgbouncer-k8s-operator"
...
 [tool.poetry.dependencies]
 python = "^3.8.10"
-ops = "^2.0"
-cryptography = "^39.0.2"
-jsonschema = "^4.17.3"
-tenacity = "^8.2.2"
+ops = "2.1.1"
+cryptography = "39.0.2"
+jsonschema = "4.17.3"
+tenacity = "8.2.2"
```
dragomirp, line 19: Renovate will update the lock and requirements but won't update `pyproject.toml` if the ^ or ~ requirement still holds true. I think this is easier to keep track of, since we wouldn't have to check the poetry lock or pip requirements to see the exact version in use.

renovate.json
taurus-forever: I wish to use team, but it doesn't work well. LGTM for now.

taurus-forever (approved):
Tnx!

marceloneppel (approved):

PR #58, opened 2023-03-21
Tweak renovate

dragomirp:
## Proposal
* Don't ignore tests, so that tests charms are also updated


renovate.json
```diff
@@ -41,5 +41,14 @@
...
   ],
+  "ignorePaths": [
+    "**/node_modules/**",
+    "**/bower_components/**",
+    "**/vendor/**",
+    "**/examples/**",
+    "**/__tests__/**",
+    "**/test/**",
+    "**/__fixtures__/**"
+  ],
```
dragomirp, line 52: These are the defaults without `"**/tests/**",`. We should be OK with removing all of them if we don't want them cluttering the config

taurus-forever (changes requested):

taurus-forever (approved):

marceloneppel (approved):

PR #76, opened 2023-04-13
Fix main branch

dragomirp:
## Issue
The backend test is actually broken, not just flaky due to CI
Also updated the poetry lock, some people claim it helps with the legacy build system failure


taurus-forever (approved):
Tnx!

PR #80, opened 2023-04-17
TLS test timeout

dragomirp:
## Issue
TLS test is consistently failing on initial deploying

## Solution
Increase the timeount for the initial deployment


dragomirp:
Doesn't seem to be topical anymore.

PR #91, opened 2023-06-07
[DPE-1779] COS

dragomirp:
## Issue
Add COS integration

## Solution
* Add an exporter service
* Add a graphana dashboard definition
* Add COS relations

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,9 +10,12 @@ import os
 import socket
 from typing import Optional
 
+from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
+from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer
 from charms.pgbouncer_k8s.v0 import pgb
 from charms.pgbouncer_k8s.v0.pgb import PgbConfig
 from charms.postgresql_k8s.v0.postgresql_tls import PostgreSQLTLS
+from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from ops.charm import CharmBase, ConfigChangedEvent, PebbleReadyEvent
 from ops.framework import StoredState
 from ops.main import main
@@ -23,11 +26,14 @@ from constants import (
     AUTH_FILE_PATH,
     CLIENT_RELATION_NAME,
     INI_PATH,
+    METRICS_PORT,
+    MONITORING_PASSWORD_KEY,
     PEER_RELATION_NAME,
     PG_GROUP,
     PG_USER,
     PGB,
     PGB_DIR,
+    PGBOUNCER_LOG_FILES,
     TLS_CA_FILE,
     TLS_CERT_FILE,
     TLS_KEY_FILE,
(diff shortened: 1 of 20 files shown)


src/charm.py
```diff
@@ -66,9 +72,22 @@ def __init__(self, *args):
...
         ]
+        self._metrics_service = "metrics_server"
+        self.grafana_dashboards = GrafanaDashboardProvider(self)
+        self.metrics_endpoint = MetricsEndpointProvider(
+            self,
+            jobs=[{"static_configs": [{"targets": [f"*:{METRICS_PORT}"]}]}],
+        )
+        self.loki_push = LogProxyConsumer(
+            self,
+            log_files=[f'{service["log_dir"]}/pgbouncer.log' for service in self._services],
```
dragomirp, line 87: In the future we should most probably listen to pebble instead.

src/charm.py
```diff
@@ -294,6 +324,40 @@ def reload_pgbouncer(self) -> None:
 
         self.check_pgb_running()
 
+    def _generate_monitoring_service(self, enabled: bool = True) -> Dict[str, str]:
+        if enabled:
+            stats_password = self.peers.get_secret("app", MONITORING_PASSWORD_KEY)
+            command = (
+                f'pgbouncer_exporter --web.listen-address=:{METRICS_PORT} --pgBouncer.connectionString="'
+                f'postgres://{self.backend.stats_user}:{stats_password}@localhost:{self.config["listen_port"]}/pgbouncer?sslmode=disable"'
```
dragomirp, line 332: Should we be using TLS here if enabled? It is a local service, so it shouldn't matter.

src/grafana_dashboards/pgbouncer-metrics.json
dragomirp, line 1: Copied over from the VM charm

src/relations/peers.py
```diff
@@ -222,7 +221,8 @@ def _on_changed(self, event: RelationChangedEvent):
             try:
                 # raises an error if this is fired before on_pebble_ready.
                 self.charm.reload_pgbouncer()
-            except ConnectionError:
+                self.charm.toggle_monitoring_layer(self.charm.backend.ready)
+            except (ConnectionError, ChangeError):
```
dragomirp, line 225: If the `listen_port` is changed, pebble cannot know and the monitoring service may fail before a PGB service is available on the new port.

taurus-forever (approved):
Tested as PPA and it does the job and looks great! Thank you!
BTW, there are some issues with pgbouncer removal, but not related to COS, let's report and address them separately. Good job done!

src/charm.py
marceloneppel: From what Mykola told me sometime ago, all the connections to the database should use TLS when available, but I think all the tools in the PG charm are using IP addresses on VM and hostnames on K8S, so I'm not sure whether it's needed for localhost. It looks that it's not worth locally.

marceloneppel (approved):

src/grafana_dashboards/pgbouncer-metrics.json
```diff
@@ -0,0 +1,1244 @@
...
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "timezone": "browser",
+  "title": "PgBouncer",
+  "uid": "aek-7cbd6b6f87",
+  "version": 2
```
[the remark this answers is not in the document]
dragomirp: Updated title, uid and tags. So dashes are available in parallel. Grafana can still display VM pgb in the K8S dashboards and vice versa.

PR #98, opened 2023-06-21
Internal changes to the charm shouldn't be released

dragomirp:
## Issue
Charm is published to charmhub on each push to main even if there are no changes to the charm itself

## Solution
Try to limit the amount of releases based on the content of the push


.github/workflows/release.yaml
```diff
@@ -6,6 +6,12 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - 'tests/**'
+      - 'documentation/**'
```
dragomirp, line 11: Maybe we should rename  to `docs` to have things consistent across charms?

taurus-forever (approved):

marceloneppel (approved):

PR #100, opened 2023-06-23
[DPE-2056] PGB extensions support

dragomirp:
## Issue
Pgbouncer would block when a legacy relation requests  extensions even if they are enabled in the backend Postgresql DB

## Solution
Check for extensions in the backend relation and block accordingly

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -20,3 +20,4 @@ parts:
       - pkg-config
     charm-binary-python-packages:
       - psycopg2-binary==2.9.6  # renovate
+      - psycopg[binary]==3.1.9  # renovate
--- a/renovate.json
+++ b/renovate.json
@@ -36,7 +36,7 @@
       "extractVersionTemplate": "Juju release"
     }, {
       "fileMatch": ["(^|/)([\\w-]*)charmcraft\\.ya?ml$"],
-      "matchStrings": ["- (?<depName>.*?)==(?<currentValue>.*?) +# renovate"],
+      "matchStrings": ["- (?<depName>.*?)(?:\\[.*?\\])?==(?<currentValue>.*?) +# renovate"],
       "datasourceTemplate": "pypi",
       "versioningTemplate": "loose"
     }
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -81,7 +81,8 @@ from ops.model import (
     WaitingStatus,
 )
 
-EXTENSIONS_BLOCKING_MESSAGE = "bad relation request - remote app requested extensions, which are unsupported. Please remove this relation."
+from constants import EXTENSIONS_BLOCKING_MESSAGE
+
 logger = logging.getLogger(__name__)
 
 
@@ -154,11 +155,12 @@ class DbProvides(Object):
                     "extensions": remote_app_databag["extensions"],
                 },
             )
-            if not self._check_extensions(remote_app_databag["extensions"].split(",")):
+            extensions = remote_app_databag["extensions"].split(",")
+            if not self._check_extensions(extensions):
                 # Do not allow apps requesting extensions to be installed.
                 logger.error(
-                    "ERROR - `extensions` cannot be requested through relations"
-                    " - they should be installed through a database charm config in the future"
+                    f"ERROR - `extensions` ({', '.join(extensions)}) cannot be requested through relations"
+                    " - Please enable extensions through `juju config` and add the relation again."
                 )
                 self.charm.unit.status = BlockedStatus(EXTENSIONS_BLOCKING_MESSAGE)
                 return False
--- a/src/relations/db.py
+++ b/src/relations/db.py
@@ -56,7 +56,7 @@ Some example relation data is below. All values are examples, generated in a run
 """
 
 import logging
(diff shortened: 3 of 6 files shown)


charmcraft.yaml
```diff
@@ -20,3 +20,4 @@ parts:
       - pkg-config
     charm-binary-python-packages:
       - psycopg2-binary==2.9.6  # renovate
```
dragomirp, line 22: Going forward we should consider porting the other charm libs to psycopg3

renovate.json
```diff
@@ -36,7 +36,7 @@
       "extractVersionTemplate": "Juju release"
     }, {
       "fileMatch": ["(^|/)([\\w-]*)charmcraft\\.ya?ml$"],
-      "matchStrings": ["- (?<depName>.*?)==(?<currentValue>.*?) +# renovate"],
+      "matchStrings": ["- (?<depName>.*?)(?:\\[.*?\\])?==(?<currentValue>.*?) +# renovate"],
```
dragomirp, line 39: Regex magic to ignore the optional deps

src/relations/db.py
```diff
@@ -136,6 +138,60 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+
+    def _block_on_extensions(self, relation, remote_app_databag: Dict) -> bool:
+        """Verifies that extensions are enabled or blocks the charm."""
+        if "extensions" in remote_app_databag:
+            self.update_databags(
+                relation,
+                {
+                    "extensions": remote_app_databag["extensions"],
+                },
+            )
```
dragomirp, line 157: We need to set this always to be able to clean up when breaking the relation

src/relations/db.py
```diff
@@ -414,10 +462,14 @@ def _on_relation_broken(self, broken_event: RelationBrokenEvent):
         if not self._check_backend() or None in [user, database]:
             # this relation was never created, so wait for it to be initialised before removing
             # everything.
-            logger.warning(
-                f"backend relation not yet available - deferring {self.relation_name}-relation-broken event."
-            )
-            broken_event.defer()
```
dragomirp, line 420: This caused a perpetual deferral with extensions, as the joining never completed.

charmcraft.yaml
taurus-forever: I thought we should get rid of 3rd party binaries here, either:
 * build from sources each time => slow
 * build and cache psycopg2-binary on Canonical side.

src/relations/db.py
```diff
@@ -136,6 +138,60 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+                {
+                    "extensions": remote_app_databag["extensions"],
+                },
+            )
+            extensions = remote_app_databag["extensions"].split(",")
+            if not self._check_extensions(extensions):
+                # Do not allow apps requesting extensions to be installed.
+                logger.error(
+                    f"ERROR - `extensions` ({', '.join(extensions)}) cannot be requested through relations"
+                    " - Please enable extensions through `juju config` and add the relation again."
```
taurus-forever, line 163: In this scenario:

> juju deploy postgresql, pgbouncer, indico
> juju relate postgresql<>pgbouncer<>indico # pgbouncer will be blocked due to the requested extension(s)
> juju config postgresql plugin_pg_trgm_enable=True

Q: can PGB be notified about config changes on PG? Sure, PGB will see new relation data on the next update_status, but it is too long...

taurus-forever (approved):

src/relations/db.py
dragomirp: At the moment the check will happen only on joining the relation. The intended user behaviour is to reconfigure the Postgresql charm and rerelate with Pgbouncer. I can add the check to update_status as well, but I don't think we should do more, as this is an escape hatch for users to be able to migrate away from the legacy relation and not something we intend to support indefinitely.

charmcraft.yaml
dragomirp: Building psycopg from source isn't straightforward, as it needs SO dependencies installed as well (libpq).  I think we should first switch to a single version before trying to figure out how to do it.

src/relations/db.py
```diff
@@ -136,6 +138,60 @@ def __init__(self, charm: CharmBase, admin: bool = False):
...
+                )
+                self.charm.unit.status = BlockedStatus(EXTENSIONS_BLOCKING_MESSAGE)
+                return False
+        return True
```
marceloneppel, line 167: Optional: I think we could return `True` when the charm blocks, to make the return of the method be more aligned with the method name. I got confused about the returned value in L229.

src/relations/db.py
marceloneppel: As Dragomir explained, I think is better to tell the user that the application charm needs to be relate again to PgBouncer after the extensions are enabled.

Currently, we don't have any field in the `postgresql_client` relation to tell that a plugin was enabled. On https://github.com/canonical/charm-relation-interfaces/pull/74 I started to create a field, but then Mykola told me that the `DataRequires` library should actively (through `is_postgresql_plugin_enabled `) check whether a plugin was enabled.

marceloneppel (approved):

PR #105, opened 2023-06-30
[DPE-1756] Log rotatation

dragomirp:
## Issue
No log rotation is done with PGB logs

## Solution
* Add scaffolds for using logrotate
* Add basic logrotate configuration
* Updated charm libs


templates/logrotate.j2
dragomirp, line 1: Should be updated once config is agreed upon.

templates/logrotate.j2
```diff
@@ -0,0 +1,16 @@
...
+    notifempty
+    nocompress
+    daily
+    create 0600 postgres postgres
+    dateext
+    dateformat -%Y%m%d_%H:%M.log
+    postrotate
+    {% for id in service_ids %}
+         pebble restart pgbouncer_{{ id }}
+    {% endfor %}
```
dragomirp, line 14: I don't think pebble can reload.

src/charm.py
```diff
@@ -248,7 +256,17 @@ def _pgbouncer_layer(self) -> Layer:
             A pebble configuration layer for as many charm services as there are available CPU
             cores
         """
-        pebble_services = {}
+        pebble_services = {
+            "logrotate": {
+                "command": "logrotate -f /etc/logrotate.conf",
+                "startup": "enabled",
+                "backoff-delay": "24h",
+                "backoff-factor": 1,
```
dragomirp, line 264: Poor man's cron daily. This will exec rotate on startup. If we want more granular behaviour, we should either add cron to the image or write a wrapper script.

templates/logrotate.j2
marceloneppel: Does `pebble signal` help on this?

marceloneppel (approved):

templates/logrotate.j2
dragomirp: It does. Thanks for the suggestion.

src/charm.py
```diff
@@ -241,7 +249,17 @@ def _pgbouncer_layer(self) -> Layer:
             A pebble configuration layer for as many charm services as there are available CPU
             cores
         """
-        pebble_services = {}
+        pebble_services = {
+            "logrotate": {
+                "command": "sh -c 'logrotate -v /etc/logrotate.conf; sleep 5'",
```
dragomirp, line 254: Pebble errors out if the command exits too fast (1s) and I couldn't find a way to disable that behaviour. Forcing the rotation (-f) usually took more, but then rules would be ignored, so I've added sleep to stabilise pebble's checks.

tests/integration/test_charm.py
```diff
@@ -55,10 +55,24 @@ async def test_multiple_pebble_services(ops_test: OpsTest):
...
-        if service[0] != "metrics_server":
+        if service[0] not in ["metrics_server", "logrotate"]:
             assert service[1] == "enabled"
             assert service[2] == "active"
+
+
+async def test_logrotate(ops_test: OpsTest):
+    """Verify that logs will be rotated."""
+    unit = ops_test.model.applications[PGB].units[0]
+    await run_command_on_unit(ops_test, unit.name, "logrotate -f /etc/logrotate.conf")
```
dragomirp, line 70: Forcing the rotation, since now logrotate will wait for rules now.

src/charm.py
```diff
@@ -329,7 +344,7 @@ def _generate_monitoring_service(self, enabled: bool = True) -> Dict[str, str]:
             command = "true"
             startup = "disabled"
         return {
-            "override": "merge",
+            "override": "replace",
```
taurus-forever, line 347: JFMI, how is it related to the current PR? Q: why do we switch to replace?

templates/logrotate.j2
taurus-forever: Q: why not `/bin/kill -HUP ${pgbouncer_pid}` ?

taurus-forever (approved):
LGTM. Two questions for my education. Tnx!

src/charm.py
dragomirp: It's not related, but while doing it, I noticed that with merge we append to the after list. So on every merge we duplicate the list of service dependencies,

PR #107, opened 2023-07-04
[DPE-1288] Charm active before backend relation

dragomirp:


marceloneppel (approved):

taurus-forever (approved):
re: "Charm waits for the first update status hook to block"
Please talk to Carl, he created a nice `blocked` behaviour in mysql-router(-k8s) without waiting for update status hook.

LGTM here, but IMHO we will have to gat rid of waiting for update-status.

PR #112, opened 2023-07-14
[DPE-2262] Fixed databases access

marceloneppel:


marceloneppel:
The library has already been merged with the main branch through another PR.

PR #117, opened 2023-07-26
[MISC] Remove requirements hashes

dragomirp:
## Issue
Latest charmcraft broke builds when requirements.txt has hashes

## Solution
Temporarily remove the hashes, so that we can release charms in the mean time (https://github.com/canonical/charmcraft/issues/1179)


tests/integration/relations/test_db.py
dragomirp, line 1: Test consistently failed. Switching to edge redis and increasing timeout.

taurus-forever (approved):

marceloneppel (approved):

PR #121, opened 2023-08-02
[DPE-1766] Juju 3 peer secrets

dragomirp:
## Issue
Charm doesn't use secrets for internal sensitive values

## Solution
* Use Juju secrets instead of peer data if available


src/charm.py
```diff
@@ -410,15 +420,176 @@ def get_hostname_by_unit(self, unit_name: str) -> str:
         unit_id = unit_name.split("/")[1]
         return f"{self.app.name}-{unit_id}.{self.app.name}-endpoints"
 
+    def _normalize_secret_key(self, key: str) -> str:
+        new_key = key.replace("_", "-")
+        new_key = new_key.strip("-")
+
+        return new_key
```
dragomirp, line 427: PGB is storing relation based secrets with generated names, so we cannot simply use an override mapping.

src/relations/db.py
```diff
@@ -168,9 +169,12 @@ def _block_on_extensions(self, relation, remote_app_databag: Dict) -> bool:
...
         """Get enabled extensions for a relation."""
-        for data in relation.data.values():
-            if "extensions" in data:
-                return data["extensions"].split(",")
+        try:
+            for data in relation.data.values():
+                if "extensions" in data:
+                    return data["extensions"].split(",")
+        except ModelError:
+            logger.warning(f"Unable to access relation data for relation {relation.id}")
```
dragomirp, line 177: I've occasionally got a model error here. I think the cause is that the relation data has already been purged by the time we call this.

src/relations/db.py
```diff
@@ -192,6 +196,25 @@ def _check_for_blocking_relations(self, relation_id: int) -> bool:
...
+            logger.warning(wait_str)
+            self.charm.unit.status = WaitingStatus(wait_str)
+            return False
+
+        logger.info(f"Setting up {self.relation_name} relation")
+        logger.warning(
+            f"DEPRECATION WARNING - {self.relation_name} is a legacy relation, and will be deprecated in a future release. "
+        )
+        return True
+
```
dragomirp, line 217: Split off because of the linter complaining about complexity.

.github/workflows/ci.yaml
```diff
@@ -124,8 +124,3 @@ jobs:
         run: tox run -e ${{ matrix.tox-environments }}-${{ env.libjuju }} -- -m '${{ steps.select-tests.outputs.mark_expression }}'
         env:
           CI_PACKED_CHARMS: ${{ needs.build.outputs.charms }}
-      - name: Dump logs
-        uses: canonical/charm-logdump-action@main
-        if: failure()
-        with:
-          app: pgbouncer-k8s
```
dragomirp, line 131: Doesn't seem to be doing anything useful.

src/relations/db.py
```diff
@@ -233,10 +237,11 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
         user = self._generate_username(join_event)
 
         if self.charm.unit.is_leader():
-            password = pgb.generate_password()
-            self.charm.peers.add_user(user, password)
+            if not (password := self.charm.get_secret(APP_SCOPE, user)):
+                password = pgb.generate_password()
+                self.charm.peers.add_user(user, password)
```
dragomirp, line 242: If password is already generated and set, don't regenerate it. This was causing the consistent mattermost failures.

marceloneppel (approved):

taurus-forever (approved):

PR #144, opened 2023-09-20
[MISC] Pluggy build failure

dragomirp:


dragomirp:
Doesn't seem to be necessary any more.

PR #147, opened 2023-09-29
[DPE-1771] Minor version upgrades

dragomirp:
Minor version upgrades for PGB.

*Since the upgrade process requires patching K8S resources the charm will require trust now*


tests/integration/relations/pgbouncer_provider/test_pgbouncer_provider.py
```diff
@@ -73,6 +73,7 @@ async def test_database_relation_with_charm_libraries(ops_test: OpsTest, pgb_cha
             application_name=PGB,
             num_units=2,
             series=CHARM_SERIES,
+            trust=True,
```
dragomirp, line 76: Not strictly necessary yet, but it will become so when we start patching the ports.

tests/integration/relations/test_db.py
```diff
@@ -213,6 +215,7 @@ async def test_relation_with_indico(ops_test: OpsTest):
     )
 
 
+@pytest.mark.unstable
 async def test_relation_with_openldap(ops_test: OpsTest):
```
dragomirp, line 219: The openldap consistently fails, in other PRs as well. Created [DPE-2685](https://warthogs.atlassian.net/browse/DPE-2685) to investigate.

tests/integration/test_upgrade.py
dragomirp, line 1: Simple sanity check test.

src/upgrade.py
```diff
@@ -0,0 +1,138 @@
...
+    def _set_rolling_update_partition(self, partition: int) -> None:
+        """Set the rolling update partition to a specific value."""
+        try:
+            patch = {"spec": {"updateStrategy": {"rollingUpdate": {"partition": partition}}}}
+            Client().patch(
+                StatefulSet,
+                name=self.charm.model.app.name,
+                namespace=self.charm.model.name,
+                obj=patch,
+            )
```
dragomirp, line 131: This requires `--trust`

src/upgrade.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Called by all units during :meth:`_on_pre_upgrade_check_action`.
+
+        Raises:
+            :class:`ClusterNotReadyError`: if cluster is not ready to upgrade.
+        """
+        self._cluster_checks()
+
+        try:
+            self._set_rolling_update_partition(self.charm.app.planned_units() - 1)
```
marceloneppel, line 79: Just a question: doesn't this trigger the upgrade in the last two units before it's needed to call the `resume-upgrade` action?

marceloneppel (approved):
Looks great!

src/upgrade.py
dragomirp: It's a copy paste from the [Postgresql charm](https://github.com/canonical/postgresql-k8s-operator/blob/main/src/upgrade.py#L220). I haven't noticed it while testing and the integration test seems to upgrade the first unit only after `resume-upgrade` was issued.

actions.yaml
```diff
@@ -7,3 +7,9 @@ set-tls-private-key:
     private-key:
       type: string
       description: The content of private key for communications with clients. Content will be auto-generated if this option is not specified.
+
+pre-upgrade-check:
+  description: Run necessary pre-upgrade checks before executing a charm upgrade.
+
+resume-upgrade:
```
delgod, line 14: In general, pgbouncer may not need `resume-upgrade` action at all.
We have it for databases because the new version of the database can change files in the data-dir in an incompatible way.
pgbouncer most probably supports downgrade without any issues, so `resume-upgrade` maybe not needed

tests/integration/relations/test_db.py
taurus-forever: IMHO, it was a Canonical registry migration, should no longer be reproducible. :crossed_fingers:

actions.yaml
taurus-forever: I would keep it for:
* identical UX for postgresql and pgbouncer
* manual ability to upgrade postgresql and pgbouncer in parallel: 
```
juju refresh pgbouncer-k8s
juju refresh postgresql-k8s
# check biggest ordinal upgrade results
juju run pgbouncer-k8s resume-upgrade
juju wait-for-active application pgbouncer-k8s
juju run postgresql-k8s resume-upgrade
juju wait-for-active application postgresql-k8s
```

taurus-forever (approved):
Nice! THANK YOU!

PR #163, opened 2023-11-22
[DPE-2844] Split indico and extension blocking tests

dragomirp:
## Issue
* Indico switched to data_interfaces
* No other tests test blocking on extensions in legacy interface

## Solution
Port the split test from PG


marceloneppel (approved):

taurus-forever (approved):
Nice!

PR #171, opened 2024-01-09
[MISC] Fix release libraries

dragomirp:
## Issue
Charmcraft strict dependencies break release libraries action

## Solution
* Regenerate `requirements.txt` the same way as for the check lib workflow
* Update dpw to 9.0.2

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,6 +33,17 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Set up environment
+        run: |
+          pipx install tox
+          pipx install poetry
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yaml
```diff
@@ -33,6 +33,17 @@ jobs:
...
+        run: |
+          pipx install tox
+          pipx install poetry
+          pipx inject poetry poetry-plugin-export
+          # TODO: Remove after https://github.com/python-poetry/poetry/pull/5980 is closed
+          poetry config warnings.export false
+      - name: Generate requirements
+        run: |
+          tox run -e build-wrapper
+          mv requirements-last-build.txt requirements.txt
```
dragomirp, line 46: Fails the same way as the check libs workflow.

PR #178, opened 2024-01-12
[DPE-3184] Secrets

dragomirp:
## Issue
PGB peer secrets implementation lagged behind the common implementation  in other charms.

## Solution
Updated the shared secrets to be in line with the Postgresql VM charm.


src/charm.py
```diff
@@ -68,7 +68,31 @@ class PgBouncerK8sCharm(CharmBase):
...
 
-        self.secrets = {APP_SCOPE: {}, UNIT_SCOPE: {}}
+        self.peer_relation_app = DataPeer(
+            self,
+            relation_name=PEER_RELATION_NAME,
+            additional_secret_fields=[
+                "monitoring-password",
+                "operator-password",
+                "replication-password",
+                "rewind-password",
```
marceloneppel, line 78: Does this operator have all these passwords?

dragomirp: Not really. Will check tomorrow what needs to be kept and what doesn't.

taurus-forever (approved):
LGTM. Please clean unnecessary secret requests as reported by Marcelo. 
Nice work!

src/charm.py
dragomirp: Pgbouncer needs arbitrary secret names, which are currently not supported by the lib. The PR will have to wait until there is a way to do this or when we know we need to change this.

src/relations/db.py
```diff
@@ -237,11 +240,7 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
         user = self._generate_username(join_event)
 
         if self.charm.unit.is_leader():
-            if not (password := self.charm.get_secret(APP_SCOPE, user)):
-                password = pgb.generate_password()
-                self.charm.peers.add_user(user, password)
-        else:
-            password = self.charm.get_secret(APP_SCOPE, user)
```
dragomirp, line 244: @delgod suggested removing the arbitrary secrets. The charm seems to function mostly the same.

tests/integration/relations/test_db.py
```diff
@@ -178,10 +179,9 @@ async def test_extensions_blocking(ops_test: OpsTest) -> None:
     await ops_test.model.add_relation(f"{PGB}:db", f"{CLIENT_APP_NAME}:db")
 
     logger.info("Wait for PGB to block due to extensions")
-    await ops_test.model.wait_for_idle(apps=[PGB], status="blocked", timeout=1000)
-    assert (
-        ops_test.model.applications[PGB].units[0].workload_status_message
-        == EXTENSIONS_BLOCKING_MESSAGE
+    leader_unit = await get_leader_unit(ops_test, PGB)
+    await ops_test.model.block_until(
+        lambda: leader_unit.workload_status_message == EXTENSIONS_BLOCKING_MESSAGE, timeout=1000
```
dragomirp, line 184: This is the main divergence I found from the previous behaviour. Since only the leader evaluates the the legacy `_on_relation_joined()` now, only the leader will block.

src/charm.py
dragomirp: List should be up to date now.

taurus-forever (approved):
Tnx!

juditnovak (approved):

marceloneppel (approved):

PR #179, opened 2024-01-16
[DPE-3062] Removing binary deps

dragomirp:
Remove charmcraft binary python packages and instead bundle necessary SO libraries to build from source.


charmcraft.yaml
```diff
@@ -26,6 +26,14 @@ parts:
       - rustc
       - cargo
       - pkg-config
-    charm-binary-python-packages:
-      - psycopg2-binary
-      - psycopg-binary
+      - libpq-dev
+  libpq:
+    build-packages:
+      - libpq-dev
```
dragomirp, line 32: I think it will be installed by the other plugin but might as well.

charmcraft.yaml
```diff
@@ -26,6 +26,14 @@ parts:
...
+    plugin: dump
+    source: /usr/lib/
+    source-type: local
+    prime:
+      - binlibs/
+    organize:
+      "*-linux-gnu/libpq.so*": binlibs/
```
dragomirp, line 39: We only want libpq SOs

src/charm.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env -S LD_LIBRARY_PATH=binlibs python3
```
dragomirp, line 1: This is magical, but seems to also work in focal. Alternatively, we can try to override the hooks to also add this, but I'm not sure there is a clean way to do that either.

tox.ini
```diff
@@ -69,7 +69,7 @@ commands =
 [testenv:unit]
 description = Run unit tests
 commands_pre =
-    poetry install --only main,charm-libs,unit --no-root
+    poetry install --only charm-libs,unit --no-root
```
dragomirp, line 72: Had issues installing psycopg3 locally for the unit tests, may end up not being necessary.

src/charm.py
marceloneppel: IMHO, it's cleaner to keep it here.

marceloneppel (commented):
The solution looks very clean. I liked it.

taurus-forever:
In general LGTM, we can try LD_LIBRARY_PATH approach.
What is the building time difference?

dragomirp:
> What is the building time difference?

The build without cached psycopg on this PR took ~4 mins. I don't think it will be drastically affected.

charmcraft.yaml
```diff
@@ -26,6 +26,14 @@ parts:
...
+  libpq:
+    build-packages:
+      - libpq-dev
+    plugin: dump
+    source: /usr/lib/
+    source-type: local
+    prime:
+      - lib/
+    organize:
+      "*-linux-gnu/libpq.so*": lib/
```
dragomirp, line 39: @delgod requested moving the SO to lib.

pyproject.toml
```diff
@@ -22,8 +22,14 @@ poetry-core = "^1.8.1"
...
-psycopg2-binary = "^2.9.9"
-psycopg = {extras = ["binary"], version = "^3.1.17"}
+
+
+[tool.poetry.group.cdeps]
+optional = true
+
+[tool.poetry.group.cdeps.dependencies]
+psycopg2 = "^2.9.9"
+psycopg = {extras = ["c"], version = "^3.1.17"}
```
dragomirp, line 32: If installing psycopg in the main group. It will be compiled at each unit test run and the unit tests won't be able to run on systems that do not have libpq installed.

taurus-forever (approved):
Hey! This is superb!!! Well done!

pyproject.toml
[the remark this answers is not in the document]
dragomirp: Libpq was not present for sure on the self-hosted runner images nor is it a given that unit tests would be run on Ubuntu flavour that has libpq. Using compiled psycopg also means that CI unit tests would have to recompile on each run. @marceloneppel @taurus-forever what do you think?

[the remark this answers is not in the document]
dragomirp: In this version unit tests task passes in ~10s: https://github.com/canonical/pgbouncer-k8s-operator/actions/runs/7714293497/job/21026204594?pr=179
If using just psycopg2 from the main group unit tests task takes ~30s: https://github.com/canonical/pgbouncer-k8s-operator/actions/runs/7714361605/job/21026430075?pr=203

Considering that the difference is up to declaring dependencies differently, I think this is the better approach.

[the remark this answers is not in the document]
taurus-forever: Removing pre-compiled binary here has a way bigger strategic effect comparing to 20 seconds saving.
BTW, 20 seconds is a noticeable time, we can safe, but as a separate PR IMHO.

Please, branch this to a new bugreport and unblock this huge merge please (as both options are safe and does the job). Tnx!

PR #186, opened 2024-01-23
[DPE-3049] templated config

dragomirp:


templates/pgb_config.j2
```diff
@@ -0,0 +1,31 @@
...
+max_client_conn = 10000
+ignore_startup_parameters = extra_float_digits,options
+server_tls_sslmode = prefer
+so_reuseport = 1
+unix_socket_dir = {{ socket_dir }}
+pool_mode = {{ pool_mode }}
+max_db_connections = {{ max_db_connections }}
+default_pool_size = 4
+min_pool_size = 2
+reserve_pool_size = 2
```
marceloneppel, line 23: Aren't those values [dynamic](https://github.com/canonical/pgbouncer-k8s-operator/blob/0220e2129d617f2311f5a649da8f9066692bf75c/lib/charms/pgbouncer_k8s/v0/pgb.py#L353)?

src/charm.py
```diff
@@ -691,19 +632,94 @@ def delete_file(self, path):
...
+                }
+                if r_hosts:
+                    pgb_dbs[f"{name}_readonly"] = {
+                        "host": r_hosts,
+                        "dbname": name,
+                        "port": r_port,
+                        "auth_user": self.backend.auth_user,
+                    }
+                    if database["legacy"]:
+                        pgb_dbs[f"{name}_standby"] = pgb_dbs[f"{name}_readonly"]
```
marceloneppel, line 719: Is `_on_changed ` being removed in this case?

dragomirp: Backend dbs will retrigger their own rerenders ATM. If we want to pass through Peers `_on_changed` we can raise flags from the backend db, but then we'll either have to put something random in the flag value or rerender twice (once to raise the flag and a second time to lower it).

templates/pgb_config.j2
dragomirp: Good catch. Should be calculated now.

lib/charms/pgbouncer_k8s/v0/pgb.py
dragomirp, line 1: Do we want to salvage anything else from here?

src/charm.py
```diff
@@ -691,19 +640,112 @@ def delete_file(self, path):
...
+                        result = {}
+                        i = 1
+                        for database in dict(databases):
+                            if database.endswith("_standby") or database.endswith("_readonly"):
+                                continue
+                            result[str(i)] = {"name": database, "legacy": False}
+                            i += 1
+                        return result
+                except Exception:
+                    logger.exception("Unable to parse legacy config format")
```
dragomirp, line 670: Try to regen the old cfg, so that we can upgrade without downtime.

templates/pgb_config.j2
```diff
@@ -0,0 +1,31 @@
+[databases]
+{% for name, database in databases.items() %}
+{{ name }} = host={{ database.host }} dbname={{ database.dbname }} port={{ database.port }} auth_user={{ database.auth_user }}
+{% endfor %}
```
dragomirp, line 4: We should be able to split out most (all?) of the moving parts and use `%include` instead to reduce the amount of rerenders. Created [DPE-3549](https://warthogs.atlassian.net/browse/DPE-3549) to track.

tests/integration/helpers/helpers.py
```diff
@@ -112,10 +112,23 @@ async def get_backend_user_pass(ops_test, backend_relation):
...
+
+    cfg = dict(parser)
+    # Convert Section objects to dictionaries, so they can hold dictionaries themselves.
+    for section, data in cfg.items():
+        cfg[section] = dict(data)
+
+    # ConfigParser object creates a DEFAULT section of an .ini file, which we don't need.
+    del cfg["DEFAULT"]
+
+    return cfg
```
dragomirp, line 131: Copied from the lib.

src/relations/peers.py
```diff
@@ -198,36 +150,17 @@ def _on_changed(self, event: HookEvent):
...
-                # raises an error if this is fired before on_pebble_ready.
-                self.charm.reload_pgbouncer()
-                self.charm.toggle_monitoring_layer(self.charm.backend.ready)
-            except (ConnectionError, ChangeError):
-                logger.error(
-                    "failed to reload pgbouncer - deferring change_event and waiting for pebble."
-                )
-                event.defer()
+        self.charm.render_pgb_config(reload_pgbouncer=True)
+        self.charm.toggle_monitoring_layer(self.charm.backend.ready)
```
dragomirp, line 163: Most of the cfg updates and restarts will happen here, when `pgb_dbs_config` gets updated.

src/relations/db.py
```diff
@@ -316,7 +304,7 @@ def _on_relation_changed(self, change_event: RelationChangedEvent):
             change_event.defer()
             return
 
-        self.update_postgres_endpoints(change_event.relation, reload_pgbouncer=True)
+        self.charm.render_pgb_config(reload_pgbouncer=True)
```
dragomirp, line 307: Relation errors out without this reload

src/charm.py
```diff
@@ -691,19 +640,112 @@ def delete_file(self, path):
...
-            FileNotFoundError when the config can't be found at INI_PATH, such as if this is called
-            before the charm has started.
-        """
-        return pgb.PgbConfig(self._read_file(INI_PATH))
+        databases = {}
+        for relation in self.model.relations.get("db", []):
+            database = self.legacy_db_relation.get_databags(relation)[0].get("database")
+            if database:
+                databases[relation.id] = {
+                    "name": database,
```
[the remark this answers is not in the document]
taurus-forever: IMHO, it is not a sensitive info (normally) but can go through secrets as well.
@delgod  ?

taurus-forever (approved):
Very interesting approach. Happe to test it in edge. Do you mind to expand unit test they are a bit shrinked here. Keep the final wording to Marcelo.

src/charm.py
```diff
@@ -486,11 +458,10 @@ def toggle_monitoring_layer(self, enabled: bool) -> None:
 
     def check_pgb_running(self):
         """Checks that pgbouncer pebble service is running, and updates status accordingly."""
-        pgb_container_unavailable = "PgBouncer container currently unavailable"
         pgb_container = self.unit.get_container(PGB)
         if not pgb_container.can_connect():
-            self.unit.status = BlockedStatus(pgb_container_unavailable)
-            logger.error(pgb_container_unavailable)
+            self.unit.status = BlockedStatus(CONTAINER_UNAVAILABLE_MESSAGE)
+            logger.warning(CONTAINER_UNAVAILABLE_MESSAGE)
```
marceloneppel, line 464: Maybe a waiting status instead of blocked status?

marceloneppel (approved):

src/charm.py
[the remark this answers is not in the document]
delgod: I would not normally put the database name into secret.

PR #191, opened 2024-01-26
[DPE-3415] Update TLS libs

dragomirp:
Update TLS libs and use new tls operator for juju 3


taurus-forever (approved):
LGTM

src/charm.py
```diff
@@ -882,7 +882,7 @@ def update_client_connection_info(self, port: Optional[str] = None):
         TODO rename
         """
         # Skip updates if backend.postgres doesn't exist yet.
-        if not self.backend.postgres:
+        if not self.backend.postgres or not self.unit.is_leader():
```
marceloneppel, line 885: Why did you add the check about the unit being the leader here and at `src/relations/db.py`?

dragomirp: Causes a lot of bogus error messages like:
`unit-pgbouncer-k8s-0: 18:40:39 ERROR unit.pgbouncer-k8s/0.juju-log This operation (update_relation_data()) can only be performed by the leader unit`, since the relation data can only be changed by the leader.

marceloneppel (approved):
LGTM!

PR #196, opened 2024-01-26
[MISC] Copy 3.1.7 fix from Mongo

dragomirp:
Copy Juju 3.1.7 fix from the mongo charm, so that we can deploy without downgrading the agent.


juditnovak (approved):
Same as https://github.com/canonical/mongodb-operator/pull/331, should B good :-)

taurus-forever (approved):

marceloneppel (approved):

PR #252, opened 2024-03-20
[MISC] Bugfixes

dragomirp:
Fixing bugs noticed during switching the VM charm to templates.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -374,15 +374,15 @@ class PgBouncerK8sCharm(CharmBase):
 
     def update_status(self):
         """Health check to update pgbouncer status based on charm state."""
+        if self.unit.status.message == EXTENSIONS_BLOCKING_MESSAGE:
+            return
+
         if self.backend.postgres is None:
             self.unit.status = BlockedStatus("waiting for backend database relation to initialise")
             return
 
         if not self.backend.ready:
             self.unit.status = BlockedStatus("backend database relation not ready")
-            return
-
-        if self.unit.status.message == EXTENSIONS_BLOCKING_MESSAGE:
             return
 
         try:
--- a/src/charm.py
+++ b/src/charm.py
@@ -457,7 +457,8 @@ class PgBouncerK8sCharm(CharmBase):
         """Checks that pgbouncer pebble service is running, and updates status accordingly."""
         pgb_container = self.unit.get_container(PGB)
         if not pgb_container.can_connect():
-            self.unit.status = WaitingStatus(CONTAINER_UNAVAILABLE_MESSAGE)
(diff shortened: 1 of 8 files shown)


tests/integration/relations/test_db.py
```diff
@@ -198,15 +197,3 @@ async def test_extensions_blocking(ops_test: OpsTest) -> None:
         raise_on_blocked=False,
         timeout=3000,
     )
-
-
-@pytest.mark.group(1)
-@pytest.mark.unstable
-async def test_relation_with_openldap(ops_test: OpsTest):
```
dragomirp, line 205: The image for the charm cannot be downloaded.

src/relations/db.py
```diff
@@ -237,7 +237,7 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
             return
 
         dbs = self.charm.generate_relation_databases()
-        dbs[join_event.relation.id] = {"name": database, "legacy": True}
+        dbs[str(join_event.relation.id)] = {"name": database, "legacy": True}
```
dragomirp, line 240: JSON keys are string

src/charm.py
```diff
@@ -644,22 +646,20 @@ def generate_relation_databases(self) -> Dict[str, Dict[str, Union[str, bool]]]:
         """Generates a mapping between relation and database and sets it in the app databag."""
         if not self.unit.is_leader():
             return {}
-        if dbs := self.get_relation_databases():
-            return dbs
```
dragomirp, line 648: Regenerating should be fast enough and if the mapping is wrong for whatever reason, always using the cache will never fix it.

src/relations/backend_database.py
```diff
@@ -335,7 +335,6 @@ def _on_relation_departed(self, event: RelationDepartedEvent):
             return
 
         self.postgres.delete_user(self.auth_user)
-        self.charm.peers.remove_user(self.auth_user)
```
dragomirp, line 338: Doesn't exist any more

tests/integration/relations/test_db.py
marceloneppel: Which image? I deployed OpenLDAP through `juju deploy openldap-charmers-openldap --channel edge` and got it in the `Waiting for database relation` state.

marceloneppel (approved):

taurus-forever (approved):
Nice stabilisation!

tests/integration/relations/test_db.py
dragomirp: Looking deeper, it looks like this is the podspec charm and the pull failure seems to originate from [the image](https://git.launchpad.net/charm-k8s-openldap/tree/src/charm.py#n55) pulled by the podspec.

PR #259, opened 2024-04-02
[MISC] Renovate tweaks and enable codecov

dragomirp:
Update renovate config and enable codecov

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,6 +37,8 @@ jobs:
           pipx install poetry
       - name: Run tests
         run: tox run -e unit
+      - name: Upload Coverage to Codecov
+        uses: codecov/codecov-action@v3
 
   build:
     name: Build charm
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -37,6 +37,8 @@ jobs:
           pipx install poetry
       - name: Run tests
         run: tox run -e unit
+      - name: Upload Coverage to Codecov
+        uses: codecov/codecov-action@v3
```
dragomirp, line 41: There should be a token generated by Is now.

marceloneppel (approved):

.github/workflows/ci.yaml
taurus-forever: I think the token is necessary for v4+ only. Right?

taurus-forever (approved):

.github/workflows/ci.yaml
dragomirp: Yes.

PR #264, opened 2024-04-04
[DPE-3777] Add node port support

phvalguima:
The node port is created if at least one single unit requests it using `external-node-connectivity=True` via `database` relation. Only applications that submit that flag as `True` receive the nodeport information. Others will receive the internal ClusterIP.

The charm internally checks at `update_connection_info()` if a given relation needs the NodePort, and updates the service accordingly with `patch_port(True)`.

Periodically, the charm reviews if node port is still necessary. runs: `patch_port` with `external_connectivity` output. If all relations are either going away or do not have `external-node-connectivity=True`, then node port is revoked. This logic is called on all main events changing the state of the charm.

### DIFF
--- a/src/relations/pgbouncer_provider.py
+++ b/src/relations/pgbouncer_provider.py
@@ -181,9 +181,6 @@ class PgBouncerProvider(Object):
         if event.departing_unit == self.charm.unit:
             self.charm.peers.unit_databag.update({self._depart_flag(event.relation): "true"})
 
-        # Check if we need to close the node port
-        self.charm.patch_port(self.external_connectivity(event))
-
     def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
         """Remove the user created for this relation, and revoke connection permissions."""
         self.update_connection_info(event.relation)


tests/integration/test_node_port.py
```diff
@@ -0,0 +1,163 @@
...
+        async with ops_test.fast_forward():
+            await ops_test.model.deploy(
+                pgb_charm,
+                resources={
+                    "pgbouncer-image": PGB_METADATA["resources"]["pgbouncer-image"][
+                        "upstream-source"
+                    ]
+                },
+                application_name=PGB,
+                num_units=1,
```
taurus-forever, line 68: As discussed in private, IMHO: we must test 2-3 units here. As in production we have 2+ PGB units for HA. 

I believe each client app have to deploy independent mysql-router (for VMs for sure... but I would keep this requirements for K8s as well). Example: pgb normally should be deployen in app mode, so wordpress in model A will have pgb in the same model, but data-integrator will in the DB model OR separate (but outside of model A => different PGB in the real deployment).

Sure, it theory the same PGB should work for 2+ apps... but create a separate bugreport if it is broken here... and we should re-check legacy interface behavior here (app1:psql + app2:postgresql_client => BOOM?).

taurus-forever (approved):
LGTM, but up to @dragomirp here!

src/relations/pgbouncer_provider.py
```diff
@@ -161,6 +181,9 @@ def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
         if event.departing_unit == self.charm.unit:
             self.charm.peers.unit_databag.update({self._depart_flag(event.relation): "true"})
 
+        # Check if we need to close the node port
+        self.charm.patch_port(self.external_connectivity(event))
+
```
dragomirp, line 186: Do we need this in both departed and broken hooks?

dragomirp (approved):

src/relations/pgbouncer_provider.py
phvalguima: @dragomirp maybe not `departed`, but on broken hook, for sure. This allows us to close the node port if no relation is requesting to be exposed any longer.

PR #275, opened 2024-04-18
Fix deletion test

phvalguima:
Removes the need for `patch_port()` at `pebble-ready` event. There is no reason to check if we need to patch the port at this event, as port should be toggled to ClusterIP / NodePort in response to client relations exclusively. Also, pebble-ready happens at every restart of the pod, which per-se does not mark as a change on exposing or not the service.

Simplifying the `external_connectivity()` to make it more readable and adding extra safeguards in the relation checking methods for node-port.

Fixes the tests to consider NodePort *or* ClusterIP accordingly.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -348,7 +348,6 @@ class PgBouncerK8sCharm(CharmBase):
         self.unit.set_workload_version(self.version)
 
         self.peers.unit_databag["container_initialised"] = "True"
-        self.patch_port()
 
     @property
     def is_container_ready(self) -> bool:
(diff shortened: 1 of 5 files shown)


taurus-forever (approved):
Good to simplify the code here. Up 2 Dragomir. LGTM.

dragomirp (approved):

dragomirp:
> Good to simplify the code here. Up 2 Dragomir. LGTM.

Let's get node port out the door first, we can iterate over it later on.

PR #277, opened 2024-04-23
[MISC] Update data-platform-libs: data_interfaces to 34

lucasgameiroborges:


tests/unit/test_charm.py
```diff
@@ -473,140 +472,106 @@ def test_invalid_secret(self, scope, is_leader, _):
...
-            assert (
-                "Non-existing field 'non-existing-secret' was attempted to be removed"
-                in self._caplog.text
-            )
-
-            self.harness.charm.remove_secret("unit", "non-existing-secret")
-            assert (
-                "Non-existing field 'non-existing-secret' was attempted to be removed"
-                in self._caplog.text
-            )
```
juditnovak, line 516: This is VERY important. Please **NEVER** ever remove tests, or part of tests.

They were there for a reason, and if they start failing that's generally signaling
 1. an error
 2. a behavior change that we need to evalate.
 
In case of (1), we need to follow up (error in the charm? error in the libs?) and perhaps roll back changes, hold back libs upgrade, or file a bug (and perhaps temporarily skip the tests until the lib bug is fixed).

 On (2) we need to evaluate if impact: does it break anything within the charm? do we need to adjust the code? do we need to adjust the tests?
 
 In this case we only need to adjust the tests, so we're kinda "good". An ERROR message got turned into a DEBUG message. However we rather keep our checks, confirming that the worklfow (that we carefully verified before) continues working as expected.
 
 Any of the above may happen -- test failure is a _signal_ for us. We can't just "resolve" it to get a green pipeline. We MUST follow up what is the underlying reason, and the adequate resolution to the actual problem/unexpected_behavior, that manifested as a test "failure".

juditnovak (changes requested):
Charm code MUST be adjusted. This PR should be litreally a copy-paste of https://github.com/canonical/pgbouncer-operator/pull/207

tests/unit/test_charm.py
lucasgameiroborges: got it, I completely agree. Will update all 3 PRs accordingly, since the review applies for all 3. Thank you for taking a look :)

juditnovak: Thank U very much :-)

dragomirp (approved):

src/charm.py
```diff
@@ -592,7 +580,13 @@ def get_secret(self, scope: Scopes, key: str) -> Optional[str]:
 
         peers = self.model.get_relation(PEER_RELATION_NAME)
```
juditnovak, line 581: We should have a check here. There is an early stage in certain charm's execution when peer relation is not yet available.

```
if not peers:
    return
```

lucasgameiroborges: good catch!

marceloneppel (approved):

taurus-forever (approved):

tests/unit/test_charm.py
```diff
@@ -515,36 +516,33 @@ def test_delete_password(self):
...
     @pytest.mark.usefixtures("use_caplog")
+    @patch("charm.JujuVersion.has_secrets", new_callable=PropertyMock, return_value=True)
     def test_delete_existing_password_secrets(self, _):
         """NOTE: currently ops.testing seems to allow for non-leader to remove secrets too!"""
         with self.harness.hooks_disabled():
             self.harness.set_leader(True)
-        self.harness.charm.set_secret("app", "operator-password", "somepw")
-        self.harness.charm.remove_secret("app", "operator-password")
-        assert self.harness.charm.get_secret("app", "operator-password") is None
+        self.harness.charm.set_secret("app", "monitoring-password", "somepw")
```
juditnovak, line 525: Quick question: is `monitoring-password` indeed the thing that exists in the charm?

(Sry I forgot which charm has which keywords :sweat_smile: )

juditnovak (approved):

tests/unit/test_charm.py
lucasgameiroborges: yep!

juditnovak: Cool, thx :-)

PR #279, opened 2024-04-24
[MISC] Don't autostart metrics service if password is not yet set

dragomirp:
Tested on https://github.com/canonical/data-integrator/pull/49

Looks like the issue was indeed the metrics service.


taurus-forever (approved):
Nice catch!

marceloneppel (approved):
LGTM!

lucasgameiroborges (approved):
Nice!

PR #306, opened 2024-06-15
[MISC] Update microk8s in renovate

dragomirp:
* Update libs
* Add renovate regex for microk8s


.github/renovate.json5
taurus-forever, line 1: Can we extract this to be shared for all Data repos?
IMHO, we are growing the same config in 20+ repos.

dragomirp: In principle, yes, in practice, we should see that it works and agree on a specific way to annotate such things.

marceloneppel (approved):

PR #319, opened 2024-06-24
[DPE-4062] Improve displayed error message if charm was deployed without `--trust` flag

lucasgameiroborges:
## Issue

K8s charms that create resources in a RBAC enabled environment need to be deployed with the `--trust` flag, otherwise they don't work. If a user deploys PGB k8s charm without that flag, there is no clear message indicating the issue.

## Solution

- Adds a integration test to see if charm enters blocked status if not deployed with `--trust`.
- Adds a check in K8s resource creation functions to catch for 403 forbidden errors. ==> This check was added everywhere `lightkube.Client().get` is used. Maybe overkill, but better safe than sorry, I think?
- Adds unit tests

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -155,21 +155,31 @@ class PgBouncerK8sCharm(CharmBase):
     @property
     def _node_name(self) -> str:
         """Return the node name for this unit's pod ip."""
-        pod = lightkube.Client().get(
-            lightkube.resources.core_v1.Pod,
-            name=self.unit.name.replace("/", "-"),
-            namespace=self._namespace,
-        )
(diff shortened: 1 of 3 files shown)


src/charm.py
dragomirp, line 1: Please add unit tests for the changes. We should try to keep Codecov happy.

lucasgameiroborges: Yep I will, I'm still not done with the changes tho. It's still in draft!! :P You are too fast!

dragomirp (approved):

marceloneppel (approved):
LGTM!

PR #342, opened 2024-07-16
[DPE-4236] Nodeport upgrade

dragomirp:
The service gets overridden during upgrade and the node port configuration gets lost.

Changed to create a separate service for nodeport.


src/charm.py
dragomirp, line 1: Creating a separate service for the node port. That should persist between upgrades.

src/relations/backend_database.py
dragomirp, line 1: Saw those fail on model removal

tests/integration/relations/pgbouncer_provider/helpers.py
dragomirp, line 1: Copypaste from the VM charm

tests/integration/test_node_port.py
```diff
@@ -147,23 +148,9 @@ async def test_node_port_and_clusterip_setup(ops_test: OpsTest):
...
-    userpass = await get_juju_secret(ops_test, info["secret-user"])
-    host, nodeport = info["endpoints"].split(":")
-
-    connection_string = (
-        f"dbname='{info['database']}' user='{userpass['username']}'"
-        f" host='{host}' port='{nodeport}' "
-        f"password='{userpass['password']}' connect_timeout=10"
+    credentials = await fetch_action_get_credentials(
+        ops_test.model.applications[DATA_INTEGRATOR].units[0]
     )
```
dragomirp, line 155: Getting the creds from data-integrator should work on either Juju

tests/unit/test_charm.py
```diff
@@ -470,14 +471,54 @@ def test_raise_untrusted_error(self, _, __, _deployed_without_trust):
         self.charm._node_name
         self.charm._node_ip
         self.charm._node_port("port")
-        self.charm.patch_port()
```
dragomirp, line 473: Stopped calling get

src/charm.py
```diff
@@ -257,59 +258,69 @@ def get_node_ports(self) -> Dict[str, Any]:
...
-            client.patch(
-                res=lightkube.resources.core_v1.Service,
-                obj=service,
-                name=service.metadata.name,
-                namespace=service.metadata.namespace,
-                force=True,
-                field_manager=self.app.name,
-                patch_type=lightkube.types.PatchType.MERGE,
-            )
-            logger.debug("Patched k8s service")
```
dragomirp, line 306: Should we keep on patching the ports on the juju service?

taurus-forever (approved):
Thank you!

src/charm.py
marceloneppel: I don't think so. I remember Mykola told me that etcd is sensible to write.

marceloneppel (approved):
Looks great! Thanks!

lucasgameiroborges (approved):

phvalguima (approved):
LGTM

PR #362, opened 2024-07-29
[MISC] Switch test app interface

dragomirp:
Switch test-app interface from `first-database` to `database`

Update libs

### DIFF
--- a/tests/integration/test_node_port.py
+++ b/tests/integration/test_node_port.py
@@ -155,7 +155,7 @@ async def test_node_port_and_clusterip_setup(ops_test: OpsTest):
             endpoint = await get_endpoint_info(ops_test, f"{app}/0", "database")
             assert "svc.cluster.local" in endpoint
 
-    await stop_continuous_writes(ops_test)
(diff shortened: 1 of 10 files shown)


marceloneppel (approved):

tests/integration/test_node_port.py
```diff
@@ -132,15 +132,14 @@ async def test_build_and_deploy(ops_test: OpsTest, pgb_charm):
 async def test_node_port_and_clusterip_setup(ops_test: OpsTest):
     """Test the nodeport."""
     # Test the writes to the database using the client app
-    psql_app = ops_test.model.applications.get(POSTGRESQL_APP_NAME)
-    await start_continuous_writes(ops_test, psql_app.name)
+    await start_continuous_writes(ops_test, PGB)
```
dragomirp, line 135: Interface has limit so can't also connect to PG. If we want the continuous_writes to go directly to PG instead of through PGB, we should add another test app.

taurus-forever: deploy second app for node port. test app is light.

taurus-forever (approved):

taurus-forever:
THANK YOU!

PR #376, opened 2024-08-16
[DPE-5145] Switch to charmed-pgbouncer-rock

dragomirp:
* Switch to charmed-pgbouncer-rock
* Update charm libs


taurus-forever (approved):
Nice! Thank you!
NodePort will be a way more lighter now! And you have opened a way to Ingress! 
Tnx!

dragomirp:
> NodePort will be a way more lighter now!

There are 300 MB of unpacked locales we can most probably remove as well. Still investigating.

PR #395, opened 2024-08-26
[DPE-5205] Port PVC error test workaround and nodeport fixes

dragomirp:
* Update libs
* Fixes to the node port logic
* Add PVC error workaround


taurus-forever (approved):
LGTM, it would be great to include this in the upcoming stable.

lucasgameiroborges (approved):
LGTM, thanks for including the `raise_on_error` workarounds here too :)

PR #405, opened 2024-09-02
[DPE-5324] Increase ruff rules

dragomirp:
Add more linting rules and change and suppress the reported issues. Added rules are:
* [flake8-bugbear](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b)
* [ruff specific rules](https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf)
* [flake8-bandit](https://docs.astral.sh/ruff/rules/#flake8-bandit-s)
* [flake8-simplify](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim)
* [pyupdate](https://docs.astral.sh/ruff/rules/#pyupgrade-up)
* [flake8-type-checking](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch)


src/constants.py
```diff
@@ -23,16 +23,17 @@
...
 CONTAINER_UNAVAILABLE_MESSAGE = "PgBouncer container currently unavailable"
 
-SECRET_LABEL = "secret"
-SECRET_INTERNAL_LABEL = "internal-secret"
-SECRET_DELETED_LABEL = "None"
+# Labels are not confidential
+SECRET_LABEL = "secret"  # noqa: S105
+MONITORING_PASSWORD_KEY = "monitoring_password"  # noqa: S105
+SECRET_INTERNAL_LABEL = "internal-secret"  # noqa: S105
+SECRET_DELETED_LABEL = "None"  # noqa: S105
```
dragomirp, line 36: Security checks will inevitably produce some false positives, so we will have to `noqa` them. I propose adding a comment on top of `noqa` blocks, so that we explain why the check is disabled.

pyproject.toml
```diff
@@ -5,7 +5,7 @@
 package-mode = false
 
 [tool.poetry.dependencies]
-python = "^3.8.10"
+python = "^3.10"
```
dragomirp, line 8: K8s charm is only running on jammy, so bump the version to be able to update coverage: https://github.com/canonical/pgbouncer-k8s-operator/pull/440

pyproject.toml
```diff
@@ -128,12 +128,19 @@ extend-ignore = [
...
+    "B011",
+    # Disable security checks for tests
+    "S",
+]
 
 [tool.ruff.lint.flake8-copyright]
 # Check for properly formatted copyright header in each file
 author = "Canonical Ltd."
 notice-rgx = "Copyright\\s\\d{4}([-,]\\d{4})*\\s+"
+min-file-size = 1
```
dragomirp, line 143: Don't expect a copyright header for empty files.

pyproject.toml
taurus-forever: It the lib/code sharing will not be the problem: ACK.

dragomirp: This is a charm specific dependency, so it shouldn't affect the lib. I think all the linting so far should be 3.8 compatible.

lib/charms/pgbouncer_k8s/v0/pgb.py
```diff
@@ -87,5 +87,6 @@
 
 def get_hashed_password(username: str, password: str) -> str:
     """Creates an md5 hashed password for the given user, in the format postgresql expects."""
-    hash_password = md5((password + username).encode()).hexdigest()
+    # Should be handled in DPE-1430
+    hash_password = md5((password + username).encode()).hexdigest()  # noqa: S324
```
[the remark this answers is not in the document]
lucasgameiroborges: so funny how it intentionally removes the `in the format postgresql expects` part

lucasgameiroborges (approved):
Nice :)

marceloneppel (approved):

taurus-forever (approved):
Thank YOU for pushing us forward!

PR #435, opened 2024-10-05
[MISC] Switch to typed charm and fixes

dragomirp:
* Switch to Pydantic configuration
* Tweaks to the legacy relation

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -16,6 +16,7 @@ from typing import Any, Dict, List, Literal, Optional, Tuple, Union, get_args
 import lightkube
 import psycopg2
 from charms.data_platform_libs.v0.data_interfaces import DataPeerData, DataPeerUnitData
+from charms.data_platform_libs.v0.data_models import TypedCharmBase
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer
 from charms.postgresql_k8s.v0.postgresql_tls import PostgreSQLTLS
@@ -23,12 +24,19 @@ from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
 from charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm
 from charms.tempo_coordinator_k8s.v0.tracing import TracingEndpointRequirer
 from jinja2 import Template
-from ops import JujuVersion, main
-from ops.charm import CharmBase, ConfigChangedEvent, PebbleReadyEvent
-from ops.framework import StoredState
-from ops.model import ActiveStatus, BlockedStatus, Relation, WaitingStatus
+from ops import (
+    ActiveStatus,
+    BlockedStatus,
+    ConfigChangedEvent,
+    JujuVersion,
+    PebbleReadyEvent,
+    Relation,
+    WaitingStatus,
+    main,
+)
 from ops.pebble import ConnectionError, Layer, ServiceStatus
 
+from config import CharmConfig
 from constants import (
     APP_SCOPE,
     AUTH_FILE_DATABAG_KEY,
@@ -79,10 +87,10 @@ Scopes = Literal[APP_SCOPE, UNIT_SCOPE]
         PgbouncerUpgrade,
     ),
 )
-class PgBouncerK8sCharm(CharmBase):
+class PgBouncerK8sCharm(TypedCharmBase):
     """A class implementing charmed PgBouncer."""
 
-    _stored = StoredState()
+    config_type = CharmConfig
 
     def __init__(self, *args):
         super().__init__(*args)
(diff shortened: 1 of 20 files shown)


tox.ini
```diff
@@ -43,7 +43,7 @@ commands =
     poetry run codespell {tox_root} --skip {tox_root}/.git --skip {tox_root}/.tox \
       --skip {tox_root}/build --skip {tox_root}/lib --skip {tox_root}/venv \
       --skip {tox_root}/.mypy_cache --skip {tox_root}/LICENSE --skip {tox_root}/poetry.lock \
-      --skip {[vars]test_charm_libs}
+      --skip {[vars]test_charm_libs} --skip {tox_root}/docs
```
dragomirp, line 46: Skipping the docs, since typos should be fixed on discourse.

src/relations/pgbouncer_provider.py
```diff
@@ -131,6 +132,27 @@ def _on_database_requested(self, event: DatabaseRequestedEvent) -> None:
...
+            # We skip the leader so we don't have to wait on the defer
+            if (
+                key != self.charm.app
+                and key != self.charm.unit
+                and data.get("pgb_dbs", "") != pgb_dbs_hash
+            ):
+                logger.debug("Not all units have synced configuration")
+                event.defer()
+                return
+
```
dragomirp, line 155: Defer until all the units are in sync (the database is added on all units). Do we want this?

src/relations/peers.py
```diff
@@ -179,4 +183,3 @@ def update_leader(self):
         """Updates leader hostname in peer databag to match this unit if it's the leader."""
         if self.charm.unit.is_leader():
             self.app_databag[LEADER_ADDRESS_KEY] = self.charm.unit_pod_hostname
-            self.charm.generate_relation_databases()
```
dragomirp, line 182: Overrides the hash while waiting for the relation to join.

src/relations/db.py
```diff
@@ -239,9 +241,23 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
...
+        for key, data in self.charm.peers.relation.data.items():
+            # We skip the leader so we don't have to wait on the defer
+            if (
+                key != self.charm.app
+                and key != self.charm.unit
+                and data.get("pgb_dbs", "") != pgb_dbs_hash
+            ):
+                logger.debug("Not all units have synced configuration")
+                join_event.defer()
+                return
```
dragomirp, line 259: Wait for all the units to be in sync (database is added). Do we want this?

src/relations/db.py
```diff
@@ -239,9 +241,23 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
         dbs = self.charm.generate_relation_databases()
         dbs[str(join_event.relation.id)] = {"name": database, "legacy": True}
         if self.admin:
-            dbs["*"] = {"name": "*", "auth_dbname": database}
+            dbs["*"] = {"name": "*", "auth_dbname": database, "legacy": False}
```
dragomirp, line 244: Legacy flag is added for backwards compatibility, adding it from the get go so that the hash remains consistent.

src/relations/db.py
```diff
@@ -229,7 +230,8 @@ def _on_relation_joined(self, join_event: RelationJoinedEvent):
         database = remote_app_databag.get("database")
         user = self._generate_username(join_event.relation)
 
-        password = pgb.generate_password()
+        databag = self.get_databags(join_event.relation)[0]
+        password = databag.get("password", pgb.generate_password())
```
dragomirp, line 234: Reuse the password if already set.

src/charm.py
```diff
@@ -79,10 +87,10 @@
         PgbouncerUpgrade,
     ),
 )
-class PgBouncerK8sCharm(CharmBase):
+class PgBouncerK8sCharm(TypedCharmBase):
     """A class implementing charmed PgBouncer."""
 
-    _stored = StoredState()
```
dragomirp, line 85: Not used for anything.

src/relations/db.py
taurus-forever: Does git blame help? I do not have the answer here.

src/relations/pgbouncer_provider.py
taurus-forever: It is to avoid DB start using until all PGB restarts happened (to avoid disconnects)?

taurus-forever (approved):
Thank you!
P.S. I still believe we need to work with IS to fix mattermost, but not mute their issue.

src/relations/pgbouncer_provider.py
dragomirp: Yes, the idea is to wait for all the units to be in sync and then update the relation data in one go. Otherwise, we either have to check when setting the read only endpoints or expect that the client will retry if the other units are still updating.

src/relations/db.py
dragomirp: It's the same case as with the new relation. We don't have to do this for PGB VM, because the subordinate relations are 1:1 and each unit is adding credentials to its own databag. Only the advertised standby unit can be out of sync, so it can only be an issue if the client starts reading from the standby immediately.

marceloneppel: I'd prefer to keep this logic for the specific case you commented:

> Only the advertised standby unit can be out of sync, so it can only be an issue if the client starts reading from the standby immediately.

marceloneppel (approved):
LGTM!

lucasgameiroborges (approved):

src/relations/db.py
lucasgameiroborges: I'd rather keep it too, better safe than sorry :)

PR #437, opened 2024-10-08
[MISC] Disable linting for docs

dragomirp:
Disable codespell for docs, since typos there will break the build and we should update those in discourse rather than fix in repo


a-velasco (approved):
Agree with disabling linting here. With the current docs setup it does not make sense for this to affect the build. It can be done separately and in the `sync-docs` branch only.

lucasgameiroborges (approved):

PR #479, opened 2025-01-06
[MISC] Disable juju 3.6 tests

dragomirp:
There are issues with TLS on juju 3.6. Disabling for now to reenable CI on main.

Updating libs


taurus-forever (approved):
LGTM, same as K8s review, please make sure we have backlog tasks for TODO here. Tnx!

marceloneppel (approved):

PR #520, opened 2025-02-06
Fix minor typo in access of ServiceSpec object

shayancanonical:
## Issue
[ServiceSpec](https://gtsystem.github.io/lightkube-models/1.32/models/core_v1/#servicespec) does not have a property `name`

## Solution
Fix access to `service.spec.type`
Also, ensure nodeport_endpoints are different than load_balancer_endpoints in test_expose_external.py

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -596,7 +596,7 @@ class PgBouncerK8sCharm(TypedCharmBase):
             return False
 
         endpoints_to_connect = [self.read_write_endpoints]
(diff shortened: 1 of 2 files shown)


lucasgameiroborges (approved):

dragomirp (approved):

taurus-forever (commented):
Wow, this is a good catch. Pending stable release until we fix this.

taurus-forever (approved):

PR #550, opened 2025-02-26
[DPE-6555] Add read only URIs

dragomirp:
- Implementing DA158: 
   - Always set a read only endpoint
   - Set read-only-uris if client has a recent data-platform-libs
- Update libs
- Don't set credentials if no endpoints

dpl changes PR: https://github.com/canonical/data-platform-libs/pull/206

### DIFF
--- a/tests/integration/relations/test_backend_database.py
+++ b/tests/integration/relations/test_backend_database.py
@@ -133,7 +133,7 @@ async def test_tls_encrypted_connection_to_postgres(ops_test: OpsTest):
         )
 
         # Relate it to the PostgreSQL to enable TLS.
-        await ops_test.model.relate(PG, tls_certificates_app_name)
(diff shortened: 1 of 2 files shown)


src/relations/peers.py
```diff
@@ -171,20 +142,9 @@ def _on_changed(self, event: HookEvent):
             self.charm.client_relation.update_endpoints()
 
     def _on_departed(self, event):
-        self.update_leader()
         self.charm.update_client_connection_info()
-        if event.departing_unit == self.leader_unit:
-            self.charm.unit.status = MaintenanceStatus(
-                "Leader unit removed - waiting for leader_elected event"
-            )
```
dragomirp, line 179: We maintain LEADER_ADDRESS_KEY in the peer data just for this status and I don't think we should be setting it anyway. Removed to bump coverage.

marceloneppel (approved):

PR #556, opened 2025-03-10
[DPE-6450] Add TLS flag and CA to relation data

dragomirp:
* Update libs
* Set and reset TLS and TLS CA relation data/secret

### DIFF
--- a/tests/integration/relations/pgbouncer_provider/helpers.py
+++ b/tests/integration/relations/pgbouncer_provider/helpers.py
@@ -141,7 +141,7 @@ async def get_tls_flags(
         return secret_data["tls"], secret_data["tls-ca"]
     else:
         return (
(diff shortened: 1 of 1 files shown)


src/relations/peers.py
```diff
@@ -172,6 +167,9 @@ def _on_changed(self, event: HookEvent):
         self.charm.toggle_monitoring_layer(self.charm.backend.ready)
         self.unit_databag["pgb_dbs"] = pgb_dbs_hash
 
+        if self.charm.unit.is_leader() and self.charm.configuration_check():
+            self.charm.client_relation.update_endpoints()
+
```
dragomirp, line 172: Changed secret/peer data may be a certificate.

taurus-forever (approved):

shayancanonical (approved):

PR #602, opened 2025-07-23
[DPE-7594] Remove pg_hba trigger

dragomirp:
* Remove pg_hba trigger
* Update charm libs
* Test fixes for blocking test app


lib/charms/postgresql_k8s/v0/postgresql.py
```diff
@@ -267,7 +267,8 @@ def create_database(
             raise PostgreSQLCreateDatabaseError() from e
 
         # Enable preset extensions
-        self.enable_disable_extensions(dict.fromkeys(plugins, True), database)
+        if plugins:
+            self.enable_disable_extensions(dict.fromkeys(plugins, True), database)
```
dragomirp, line 271: The Pgbouncer user may not have access to the database after creating it, so the connection can be rejected here. Since PGB doesn't set any extensions, we can skip.

src/relations/backend_database.py
```diff
@@ -476,3 +487,17 @@ def check_backend(self) -> bool:
             self.charm.unit.status = WaitingStatus(wait_str)
             return False
         return True
+
+    def sync_hba(self, user: str) -> None:
+        """Wait for user to appear in pg_hba table."""
+        # Check for version supporting hardening
+        if self.backend_version and self.backend_version < "14.16":
+            return
```
dragomirp, line 495: `14.15` is the current stable. Since it has no hardening, there's no point in waiting.

marceloneppel (approved):

PR #703, opened 2026-03-13
[DPE-8490] Use non-root for PostgreSQL K8s charm

taurus-forever:


taurus-forever:
Not that simple. To be continued...

PR #768, opened 2026-06-02
[MISC] Don't clean up runner disk

dragomirp:
Stop cleaning the runner disk.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -88,20 +88,11 @@ jobs:
     needs:
       - collect-integration-tests
     runs-on: ${{ matrix.job.runner }}
-    timeout-minutes: 226  # Sum of steps `timeout-minutes` + 5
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -88,20 +88,11 @@ jobs:
     needs:
       - collect-integration-tests
     runs-on: ${{ matrix.job.runner }}
-    timeout-minutes: 226  # Sum of steps `timeout-minutes` + 5
+    timeout-minutes: 216  # Sum of steps `timeout-minutes` + 5
     environment:
       name: integration-test
       deployment: false
     steps:
-      - name: Free up disk space
```
dragomirp, line 96: There should be enough space with the current layout, so there's no need to spend time removing pre-installed components.

marceloneppel (approved):

taurus-forever (approved):

PR #799, opened 2026-07-07
[MISC] Limit solver to 1d old dependencies

dragomirp:
Follow up to https://github.com/canonical/data-platform/pull/42.

Add minimum age for Poetry lock

### DIFF
--- a/poetry.toml
+++ b/poetry.toml
@@ -1,0 +1,3 @@
+[solver]
+min-release-age = 1
+min-release-age-exclude = "postgresql-charms-single-kernel"


carlcsaposs-canonical (approved):
nit: could we set this in pyproject.toml instead to keep everything in one place?

carlcsaposs-canonical:
> nit: could we set this in pyproject.toml instead to keep everything in one place?

nvm looks like that's not possible (https://github.com/orgs/python-poetry/discussions/10555#discussioncomment-14461277), which it a bit surprising

marceloneppel (approved):

Issue #740, opened 2026-04-22
Connection persists after error: DISCARD ALL cannot run inside a transaction block


trollLemon:
## Steps to reproduce

1. Deploy canonical-livepatch-server-k8s
2. Deploy postgresql-k8s and pgbounce-k8s
3. Set the requried livepatch configurations: `juju config livepatch-server server.is-hosted=true server.url-template=testing`
4. Scale up the livepatch server charm to 4 units.
5. relate postgresql to pgbouncer
6. relate livepatch-server:database to pgbouncer

## Expected behavior
Livepatch server should run schema migrations on the leader unit, and the other units should connect to postgresql via the relation. 

## Actual behavior

Livepatch server fails with `hook failed: "database-relation-changed"`, failing due to attempting to insert prepared statements with errors such as `Error: ERROR: prepared statement "lrupsc_1_0" already exists (SQLSTATE 42P05)`

Based on the logs from the postgres unit, the schema migrations run, but after the connection closes and an error occurs: `DISCARD ALL cannot run inside a transaction block`.
## Versions

Operating system: Ubuntu 22.04.5 LTS

Juju CLI: 3.6.21-genericlinux-amd64

Juju agent: 3.6.21

postgresql-k8s charm revision: 14.20
pgbouncer-k8s charm revision: 1.21.0

microk8s: MicroK8s v1.25.16 revision 6575

## Log output

Juju debug log: (see attached file)
Postgres logs from postgres container:
```
2026-04-21 15:45:25 UTC [442]: user=relation_id_7,db=livepatch-server,app=[unknown],client=10.1.215.37,line=711 LOCATION:  log_audit_event, pgaudit.c:769
2026-04-21 15:45:25 UTC [442]: user=relation_id_7,db=livepatch-server,app=[unknown],client=10.1.215.37,line=712 LOG:  00000: AUDIT: SESSION,305,1,MISC,COMMIT,,,commit,<not logged>
2026-04-21 15:45:25 UTC [442]: user=relation_id_7,db=livepatch-server,app=[unknown],client=10.1.215.37,line=713 LOCATION:  log_audit_event, pgaudit.c:769
2026-04-21 15:45:25 UTC [442]: user=relation_id_7,db=livepatch-server,app=[unknown],client=10.1.215.37,line=714 LOG:  00000: AUDIT: SESSION,306,1,MISC,BEGIN,,,begin isolation level serializable,<not logged>
[...]

PR #817, opened 2026-08-10
[DPE-9936] Conditionally add discard directives

dragomirp:
Relates to https://github.com/canonical/pgbouncer-k8s-operator/issues/740
Port of https://github.com/canonical/pgbouncer-operator/pull/711

## Issue
When adapting the charm to Postgresql 16 charm roles management, we added a workaround for recycled connections, so that the `charmed_{dbname}_owner` role is not lost. The workaround generates excessive error logs in the Postgresql charm and does not properly recycle the relation.

## Solution
The workaround is not necessary for Postgresql 14 charm, since roles work differently and for Postgresql 16 charm it should match the discard command without resetting the role.


taurus-forever (approved):
It smells a bit, but no better ideas at the moment.

marceloneppel (approved):

PR #832, opened 2026-08-27
fix(config): preserve charmed role by setting server_reset_query on PostgreSQL 16+

marceloneppel:
## Issue

On PostgreSQL 16+ (Ubuntu 24.04 Noble), default client connection recycling uses PgBouncer's built-in \`DISCARD ALL\`, which resets session authorization and drops privileges back to the relation user. Consequently, subsequent operations in the \`public\` schema fail with \`InsufficientPrivilege: permission denied for schema public\` due to PostgreSQL 15+ revoking public schema \`CREATE\` privileges from \`PUBLIC\`.

The intended custom \`server_reset_query\` was omitted due to an inverted boolean check in \`backend_major_version\` and a strict inequality \`> 16\` in the configuration template.

## Solution

Fix \`backend_major_version\` condition to check \`if not self.relation or not self.backend_version:\`, and update \`templates/pgb_config.j2\` to \`{% if backend_version >= 16 -%}\` so that the tailored reset query is emitted for PostgreSQL 16+.

Tested on https://github.com/canonical/postgresql-test-app/pull/543.

### DIFF
--- a/src/relations/backend_database.py
+++ b/src/relations/backend_database.py
@@ -165,10 +165,10 @@ class BackendDatabaseRequires(Object):
             return version
         return ""
 
-    @cached_property
+    @property
     def backend_major_version(self) -> int:
         """Backend Postgresql version."""
-        if not self.relation or self.backend_version:
+        if not self.relation or not self.backend_version:
             return 0
 
         try:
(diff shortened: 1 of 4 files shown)


marceloneppel:
Superseded by https://github.com/canonical/pgbouncer-k8s-operator/pull/833.

PR #833, opened 2026-08-27
[DPE-9936] Fix template rendering

dragomirp:
Bugfix of the bugfix. Broke the test application CI.

Correctly check the backend version and render discard statement for equal or greater than 16.

### DIFF
--- a/src/relations/backend_database.py
+++ b/src/relations/backend_database.py
@@ -168,7 +168,7 @@ class BackendDatabaseRequires(Object):
     @cached_property
     def backend_major_version(self) -> int:
         """Backend Postgresql version."""
-        if not self.relation or self.backend_version:
+        if not self.relation or not self.backend_version:
             return 0
 
         try:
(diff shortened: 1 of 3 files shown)


src/relations/backend_database.py
```diff
@@ -168,7 +168,7 @@ def backend_version(self) -> str:
     @cached_property
     def backend_major_version(self) -> int:
         """Backend Postgresql version."""
-        if not self.relation or self.backend_version:
+        if not self.relation or not self.backend_version:
```
dragomirp, line 171: Instead of checking was defaulting to 14.

templates/pgb_config.j2
```diff
@@ -40,6 +40,6 @@ client_tls_ca_file = {{ ca_file }}
 client_tls_cert_file = {{ cert_file }}
 client_tls_sslmode = prefer
 {% endif %}
-{% if backend_version > 16 -%}
+{% if backend_version >= 16 -%}
```
dragomirp, line 43: Should be applied for 16+.

marceloneppel (approved):

carlcsaposs-canonical (approved):

PR #836, opened 2026-08-28
[DPE-11007] Follow juju CPU constraints when sizing pgbouncer instances

### LINKED ISSUE #834: Number of Instance Assignment Doesn't Follow juju constraint setup
## Steps to reproduce

1. juju deploy pgbouncer-k8s --channel 1/stable --trust --constraints "cpu-power=2000 mem=2G arch=amd64"
2. Wait for the deployment
3. juju ssh --container pgbouncer pgbouncer-k8s/0 ps aux
4. Observe that there are 4 pgbouncer instances

## Expected behavior
Expect to have 2 pgbouncer instances according to cpu-power=2000

## Actual behavior
user@server:$ juju constraints pgb
arch=amd64 cores=2 cpu-power=2000 mem=2048M
user@server:$ juju ssh --container pgbouncer pgb/0 ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0 1236684 14232 ?       Ssl  07:04   0:00 /charm/bin/pebble run --create-dirs --hold --http :
postgres      15  0.0  0.0  21104  6300 ?        Sl   07:04   0:00 pgbouncer /var/lib/pgbouncer/instance_0/pgbouncer.i
postgres      17  0.0  0.0  21104  6220 ?        Sl   07:04   0:00 pgbouncer /var/lib/pgbouncer/instance_1/pgbouncer.i
postgres      19  0.0  0.0  21104  6068 ?        Sl   07:04   0:00 pgbouncer /var/lib/pgbouncer/instance_2/pgbouncer.i
postgres      21  0.0  0.0  21104  6076 ?        Sl   07:04   0:00 pgbouncer /var/lib/pgbouncer/instance_3/pgbouncer.i
postgres      23  0.0  0.0 1156096 10380 ?       Sl   07:04   0:00 pgbouncer_exporter --web.listen-address=:9127 --pgB
root          91  0.0  0.0   2892  1776 pts/0    Ss+  07:19   0:00 sh -c ps aux
root          98  0.0  0.0   7064  3036 pts/0    R+   07:19   0:00 ps aux

## Versions
Jaas 3.6.24
pgbouncer-k8s charm revision: 1.21.0 rev 562 

## Additional context
n/a

delgod:
Fixes #834.

The instance count came from `os.cpu_count()`, which reports the whole node and ignores
the pod's cgroup quota. A unit deployed with `--constraints "cpu-power=2000"` (`cores=2`)
ran 4 instances instead of 2.

`_instance_count()` now reads the `pgbouncer` container's CPU limit (or request) from the
pod spec, falling back to the container's cgroup quota, then to the host CPU count. Still
rounded up and clamped to 2–4, so unconstrained deployments are unchanged.

#834 also proposes a `juju config instances=N` option; this PR makes the count follow the
constraints already set instead, so the issue should stay open for that request.

**Testing:** 42 unit tests covering each source and fallback. The integration test
compared against raw `nproc --all`, which only held on 2–4 core runners; it now applies
the same clamp.

**Before merge:** confirm Juju puts a `cores` constraint on the `pgbouncer` container and
not just the `charm` container — if the latter, detection falls through to the host CPU
count and #834's repro is unchanged.

### DIFF
--- a/src/cpu.py
+++ b/src/cpu.py
@@ -1,12 +1,12 @@
 # Copyright 2026 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-"""Detection of the CPU allocation provisioned for the pgbouncer container.
+"""Detection of the CPU quota Kubernetes enforces on the pgbouncer container.
 
-PgBouncer is single-threaded, so the charm runs one process per provisioned CPU. The
-helpers here read a single source each; `PgBouncerK8sCharm._instance_count` is where they
-are tried in order. Note that `os.cpu_count()`, the last resort there, reports the CPUs of
-the whole node, ignoring both cgroup quotas and affinity.
+Juju turns a `cores`/`cpu-power` constraint into a CPU *limit* on the workload container,
+which the kernel enforces as a cgroup quota. Reading that quota from inside the container
+needs no Kubernetes API call and no `juju trust`. The charm container has no CPU limit of
+its own, so `os.cpu_count()` there reports the whole node.
 """
 
 import logging
@@ -19,37 +19,6 @@ logger = logging.getLogger(__name__)
 CGROUP_V2_CPU_MAX = "/sys/fs/cgroup/cpu.max"
 CGROUP_V1_QUOTA = "/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
 CGROUP_V1_PERIOD = "/sys/fs/cgroup/cpu/cpu.cfs_period_us"
-
-MIN_INSTANCES = 2
-MAX_INSTANCES = 4
-
-
-def parse_quantity(value: str | None) -> float | None:
-    """Convert a Kubernetes CPU quantity ("2", "0.5", "1500m") into a number of cores."""
-    if not value:
-        return None
-    try:
-        if value.endswith("m"):
-            return float(value[:-1]) / 1000
-        return float(value)
-    except ValueError:
-        logger.warning("Unable to parse CPU quantity %r", value)
-        return None
-
-
-def container_cpu_limit(pod, container_name: str) -> float | None:
-    """Cores provisioned for a container in the pod spec, or None if it is unconstrained."""
-    spec = getattr(pod, "spec", None)
-    for container in getattr(spec, "containers", None) or []:
-        if container.name != container_name:
-            continue
-        resources = container.resources
-        if resources is None:
-            return None
-        for quantities in (resources.limits, resources.requests):
-            if quantities and (cores := parse_quantity(quantities.get("cpu"))):
-                return cores
-    return None
 
 
 def cgroup_cpu_limit(container: Container) -> float | None:
--- a/src/cpu.py
+++ b/src/cpu.py
@@ -1,0 +1,85 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Detection of the CPU allocation provisioned for the pgbouncer container.
+
+PgBouncer is single-threaded, so the charm runs one process per provisioned CPU. The
+helpers here read a single source each; `PgBouncerK8sCharm._instance_count` is where they
+are tried in order. Note that `os.cpu_count()`, the last resort there, reports the CPUs of
+the whole node, ignoring both cgroup quotas and affinity.
+"""
+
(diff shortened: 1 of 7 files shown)


tests/integration/test_charm.py
dragomirp, line 28:
```suggestion
            constraints={"cpu-power": 2000},
            trust=True,
```
We can try to constrain and verify that the minimum is set, which will be consistent instead of pulling the `cpu` module and `ops`.

src/cpu.py
dragomirp, line 1: I think most other charms just check the container quota (https://github.com/canonical/postgresql-k8s-operator/blob/9d4f71728b497db276299c827cb0254b0cff8454/src/charm.py#L2557-L2563 and https://github.com/canonical/postgresql-k8s-operator/blob/9d4f71728b497db276299c827cb0254b0cff8454/src/utils.py#L64-L73). Should we be checking the CGROUPs as well?

tests/integration/test_charm.py
delgod: Done in 337c478d — thanks, this is better than what I had.

src/cpu.py
delgod: `get_node_cpu_cores()` reads `node.status.allocatable["cpu"]`, which looks to me like the node's capacity rather than the container's quota. So I am not sure that postgresql-k8s reads a container limit anywhere.

 Since your comment, I've dropped the pod-spec path, so the cgroup is now the only source in 337c478d. Checked on microk8s with `--constraints "cpu-power=2000"`:
 
  |                     | value                                                       |
 |---------------------|-------------------------------------------------------------|
 | pgbouncer container | limits.cpu: 2, cpu.max = 200000 100000, nproc --all = 16    |
 | charm container     | cpu.max = max — no quota, hence os.cpu_count() = 16         |
 | processes           | rev 562 → 4, this branch → 2, unconstrained → 4 (unchanged) |
 
Reasons to avoid the API here: `_instance_count()` runs in `__init__`, so the pod-spec version made a lightkube call on every hook — in `test_peers` that was 69 API connections vs 2 on main, 17 of which timed out and killed pebble-ready 9 times. postgresql-k8s calls `get_available_resources()` from `update_config()` rather than the constructor, so it doesn't hit that. Reading the cgroup also means detection works without juju trust.

dragomirp: Checked canonical k8s with `cpu-power=2000` and `cpu-power=500`, services were limited to 2.

dragomirp:
We may additionally need to stop extra services on upgrade, like in PGB VM:
https://github.com/canonical/pgbouncer-operator/blob/dfda2c17e53f6654aa07bcc8b9267103b633748f/src/upgrade.py#L136-L140

But it can be done as a follow up.

dragomirp (dismissed):

delgod:
> We may additionally need to stop extra services on upgrade, like in PGB VM: https://github.com/canonical/pgbouncer-operator/blob/dfda2c17e53f6654aa07bcc8b9267103b633748f/src/upgrade.py#L136-L140

The K8s case differs from VM here: juju refresh rolls the StatefulSet, so the pod is recreated with the new values.

src/cpu.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Detection of the CPU quota Kubernetes enforces on the pgbouncer container.
+
+Juju turns a `cores`/`cpu-power` constraint into a CPU *limit* on the workload container,
```
marceloneppel, line 6: I'm not 100% sure, but it seems that `cores` is not supported as a constraint for K8s on Juju: https://github.com/juju/juju/blob/2160900254e57e56614d40ee5937749193886f5d/internal/provider/kubernetes/constraints.go#L12.

marceloneppel (approved):
LGTM

dragomirp (approved):

src/cpu.py
delgod: as I can see `cpu-power=2000` for 2 cores works fine

marceloneppel: Yes, I believe it works. I commented about `cores`.

